PR #5, opened 2026-05-20
Web and meta update

SaxonF:
image

Updated the site and evals to add new metadata via prompt.md front matter (vs naming conventions) and updated the design to  better handle more evals across more categories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,25 +114,14 @@ npm run eval -- --model gpt-5.4-mini
 Create a directory named:
 
 ```
-evals/<category>-<subcategory>-<NNN>-<slug>/
+evals/<short-id>/
 ```
 
 Rules:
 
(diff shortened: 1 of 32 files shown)


mattrossman (approved):
New web updates look really slick 💅

Some non-blocking notes:
- Eventually I'd like some way to identify regression tests apart benchmark tests, atm everything is part of the benchmark. Do you think we should support a `suite:` frontmatter field, or store regression in separate folder(s)?
- The frontmatter approach is handy in that metadata is visble next to the prompt, only downside is we don't get type safety/hints on the allowed values for each metadata field. Perhaps core could expose a `defineEval` function similar to `defineExperiment` that could be used in the `EVAL.ts` file to enforce allowed values. I'm also fine to stick w/ frontmatter in the prompt as it's not a major issue atm.
- There are some `npm run lint` errors in `apps/web`, maybe we should address and run that check in CI to catch early?

PR #6, opened 2026-05-28
feat: add LLM judge support for eval scoring

mattrossman:
Some scenarios (e.g. regression case [AI-593](https://linear.app/supabase/issue/AI-593/add-eval-for-unhealthy-project-recovery-answer)) involve fuzzy scoring of the agent's response which may not be feasible with determinstic string checks.

- Adds `judge()` to `@supabase-evals/core` — runs a minimal LLM judge via AI SDK (`gpt-5.5` low reasoning / low verbosity by default) against an arbitrary input string and rubric. Compose with the exported `serializeTranscript()` to judge a full agent transcript, or pass `agentReport` to judge just the final report.
- Threads structured `transcript` through the harness so scorers can inspect the full agent conversation (every step's assistant text + tool calls).
- Replaces string result `notes` / scenario scores with structured `checks[]`, including deterministic checks and LLM-backed checks with collapsible notes in the web UI.
- Updates `detect-security-001-public-table` to demonstrate the workflow with an LLM check that verifies the exposed table and proposed fix (we can revert this part if desired, it mainly exists to demo LLM checks)

To run the example eval demoing the judge:

```
npm run eval -- --experiment openai-gpt-5.4-mini --eval detect-security-001-public-table --runs 1 --force
```

Output includes checks pass rate:
```
RUN  openai-gpt-5.4-mini x detect-security-001-public-table
  -> PASS (checks 4/4, attempts 1)
```

Rendered result in the web app:

CleanShot 2026-05-29 at 15 00 40@2x

Closes [AI-758](https://linear.app/supabase/issue/AI-758/support-llm-as-a-judge-assertions-in-evalts)

### DIFF
--- a/evals/design-frontend-001-todos-app/EVAL.ts
+++ b/evals/design-frontend-001-todos-app/EVAL.ts
@@ -6,28 +6,39 @@ const scorer: ProjectScorer = async (ctx) => {
   if (!build.ok) {
     return {
       passed: false,
-      score: 0,
-      notes: `vite build failed:\n${trimOutput(build.stderr || build.stdout)}`,
+      assertions: [
+        {
+          type: "deterministic",
+          name: "vite build passed",
+          passed: false,
+          notes: trimOutput(build.stderr || build.stdout),
+        },
+      ],
     };
   }
 
   if (!vitest) {
-    return { passed: false, score: 0.25, notes: "vitest did not run" };
+    return {
+      passed: false,
+      assertions: [{ type: "deterministic", name: "vitest ran", passed: false }],
+    };
   }
 
-  const total = (vitest.passed ?? 0) + (vitest.failed ?? 0);
-  const score = total > 0 ? (vitest.passed ?? 0) / total : vitest.ok ? 1 : 0.25;
   return {
     passed: vitest.ok,
-    score,
-    notes: vitest.ok
-      ? "vite build and vitest passed"
-      : [
-          "vitest failed:",
-          ...((vitest.failures?.length ? vitest.failures : [vitest.stderr || vitest.stdout]).map(
-            trimOutput
-          )),
-        ].join("\n"),
+    assertions: [
+      { type: "deterministic", name: "vite build passed", passed: true },
+      {
+        type: "deterministic",
+        name: "vitest passed",
+        passed: vitest.ok,
+        notes: vitest.ok
+          ? undefined
+          : (vitest.failures?.length ? vitest.failures : [vitest.stderr || vitest.stdout])
+              .map(trimOutput)
+              .join("\n"),
+      },
+    ],
   };
 };
 
--- a/evals/design-frontend-001-todos-app/EVAL.ts
+++ b/evals/design-frontend-001-todos-app/EVAL.ts
@@ -6,7 +6,7 @@ const scorer: ProjectScorer = async (ctx) => {
   if (!build.ok) {
     return {
       passed: false,
-      assertions: [
+      checks: [
         {
           type: "deterministic",
           name: "vite build passed",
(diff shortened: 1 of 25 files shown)


evals/design-frontend-001-todos-app/EVAL.ts
```diff
@@ -6,28 +6,39 @@ const scorer: ProjectScorer = async (ctx) => {
...
-      score: 0,
-      notes: `vite build failed:\n${trimOutput(build.stderr || build.stdout)}`,
+      checks: [
+        {
+          type: "deterministic",
+          name: "vite build passed",
+          passed: false,
+          notes: trimOutput(build.stderr || build.stdout),
+        },
+      ],
```
gregnr: Is there a reason why we omit the vitest check here (and similarly, the vite build check in vitest check below), and only return both checks if both passed? The early exit control flow just feels a bit complex vs a single return statement with all checks that conditionally set their `passed` flags.

packages/core/src/index.ts
```diff
@@ -66,7 +68,44 @@ export type EvalResult = {
 
 export interface ScoreResult {
   passed: boolean;
-  score: number;
+  checks?: CheckResult[];
+}
+
+export type CheckResult = {
+  type: "deterministic" | "llm";
```
gregnr: What signal does this type give us downstream? Just curious if it is serving a purpose

gregnr (approved):

packages/core/src/index.ts
mattrossman: The purpose is mainly on the UI side, for `"llm"` checks it shows a bot icon to distinguish it from determinsitic checks and labels the notes dropdown differently.

CleanShot 2026-05-29 at 15 00 40@2x

CleanShot 2026-05-29 at 23 11 50@2x

Alternatively, we could put LLM notes in a separate field like `llmNotes` and infer type from its presence. Or assume that checks are deterministic by default and use an optional field like `.isLlmCheck` to opt-in the the judge UI variant.

I'm not a huge fan of having to specify `"deterministic"` on most of these checks but the distinction in the UI is nice to have I think.

evals/design-frontend-001-todos-app/EVAL.ts
mattrossman: I agree a single return at the end is clearer, and would keep the number of checks for an eval consistent across runs. I noticed this in the edge function evals too, if an error occurs it short circuits and returns a single failing check instead of showing the other checks as failing or skipped. That design precedes this PR, but I'll take the opportunity to clean it up so the checks are declared in one place at the end.

mattrossman: I cleaned up the flow to have a single return in fd67d0269b31d8f983379db7c42dda0c866c4708

The edge function scenarios like `design-functions-001-order-total` aren't as straightforward to refactor to a single return without introducing a lot of mutable variables. I think in a follow-up PR, we could introduce a small helper to make it easier to declare checks without manually managing conditional intermediate state.

So instead of:

```ts
let sawExpectedThing: boolean | undefined;
let handledBadInput: boolean | undefined;
let error: string | undefined;

try {
  const firstResult = await doSomething();
  sawExpectedThing = firstResult.ok;

  const secondResult = await doSomethingElse();
  handledBadInput = secondResult.ok;
} catch (e) {
  error = String(e);
}

return [
  {
    name: "saw expected thing",
    passed: sawExpectedThing === true,
    notes: sawExpectedThing === undefined ? error : undefined,
  },
  {
    name: "handled bad input",
    passed: handledBadInput === true,
    notes: handledBadInput === undefined ? error : undefined,
  },
];
```

it could look like:

```ts
return runChecks([
  check("saw expected thing", async () => {
    const result = await doSomething();
    return result.ok;
  }),

  check("handled bad input", async () => {
    const result = await doSomethingElse();
    return result.ok;
  }),
]);
```

packages/core/src/index.ts
mattrossman: I changed it to use `.judgeNotes` (corresponding w/ the UI label) for the UI distinction between generic & LLM judge notes, since the majority of checks are deterministic the explicit `type` was adding more diff noise than necessary.

PR #17, opened 2026-06-10
feat: add Data API empty results eval

Rodriguespn:
See "Diagnose empty Data API results" scenario from [Supa-bench (condensed)](https://app.notion.com/p/supabase/Supa-bench-condensed-3675004b775f8066824ce11e9fd06304?source=copy_link#37b5004b775f80cea8adfb03d512d2a7).

**Additions**
- Benchmark eval `resolve-dataapi-001-empty-results` for diagnosing why the Data API silently returns an empty array despite existing rows, tracing the cause to missing RLS policies on a table with RLS enabled, and fixing it with owner-scoped policies.

**To run the eval**
```bash
npm run eval -- --eval resolve-dataapi-001-empty-results --experiment claude-sonnet-4.6 --force --runs 1
```

**What the eval checks**
- RLS remains enabled on the `bookmarks` table after the fix.
- User A reads only their own bookmarks.
- User B cannot read user A's bookmarks.
- Anon reads no bookmarks.
- User A can save a new bookmark.
- User B cannot insert a bookmark as user A.
- Agent diagnosed the empty results as an RLS deny-all issue and added owner-scoped policies using `USING (user_id = auth.uid())`.

Closes AI-822

### DIFF
--- a/evals/resolve-dataapi-001-empty-results/EVAL.ts
+++ b/evals/resolve-dataapi-001-empty-results/EVAL.ts
@@ -1,0 +1,237 @@
+import {
+  judge,
+  serializeTranscript,
+  type CheckResult,
+  type SupabaseClient,
+  type ToolEvalContext,
+  type ToolScorer,
+} from "@supabase-evals/core";
+import { stripIndent } from "common-tags";
+
+const PASSWORD = "secret123";
+const BOOKMARK_A1 = "a bookmark alpha";
+const BOOKMARK_A2 = "a bookmark beta";
+const BOOKMARK_B = "b bookmark gamma";
+
+const scorer: ToolScorer = async (ctx) => {
+  try {
+    const setup = await setupTestUsers(ctx);
+    if ("failure" in setup) {
+      return { passed: false, checks: [setup.failure] };
+    }
+    const users = setup.users;
+
+    const checks: CheckResult[] = [
+      await checkRlsStillEnabled(ctx),
+      await checkUserAReadsOwnBookmarks(users),
+      await checkUserBCannotReadUserABookmarks(users),
+      await checkAnonReadsNoBookmarks(ctx),
+      await checkUserACanSaveNewBookmark(users),
+      await checkUserBCannotInsertAsUserA(ctx, users),
+      await checkRlsDiagnosisAndOwnerPolicies(ctx),
+    ];
+
+    return {
+      passed: checks.every((check) => check.passed),
+      checks,
+    };
+  } catch (error) {
+    const msg = error instanceof Error ? error.message : String(error);
+    return {
+      passed: false,
+      checks: [
+        {
+          name: "scorer evaluated Data API fix",
+          passed: false,
+          notes: msg,
+        },
+      ],
+    };
+  }
+};
+
+export default scorer;
+
+type TestUsers = {
+  clientA: SupabaseClient;
+  clientB: SupabaseClient;
+  userAId: string;
+  userBId: string;
+};
+
+async function setupTestUsers(
+  ctx: ToolEvalContext,
+): Promise<{ users: TestUsers } | { failure: CheckResult }> {
+  const clientA = ctx.client;
+  const clientB = ctx.getClient();
+
+  const { data: authA, error: authAError } = await clientA.auth.signUp({
+    email: `empty-results-a-${Date.now()}@example.com`,
+    password: PASSWORD,
+  });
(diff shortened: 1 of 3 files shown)


evals/resolve-dataapi-001-empty-results/EVAL.ts
```diff
@@ -0,0 +1,237 @@
...
+  // data/query problem, and fixed with owner-scoped policies.
+  const verdict = await judge({
+    input: serializeTranscript(ctx.transcript, {
+      includeToolCallInputs: true,
+    }),
+    rubric: stripIndent`
+      Pass if the assistant handles the empty Data API results as a Row Level Security issue and fixes it by creating RLS policies that let users read and insert only their own bookmarks.
+
+      A passing answer should:
+      - Diagnose that RLS is enabled on the bookmarks table with no policies, so the Data API returns zero rows to every user (deny-all by default).
```
Rodriguespn, line 211: Should we be as strict as the agent acknowledged that the Postgres default policy deny-all? Tested with sonnet-4.6 and it acknowledged this but no sure if other models must do the same if at the end create RLS policies and fix the problem

mattrossman: I think it's fine if it doesn't use that exact wording, as long as it generally diagnoses the missing policies + Data API behavior as the cause.

evals/resolve-dataapi-001-empty-results/PROMPT.md
```diff
@@ -0,0 +1,21 @@
+---
+stage: resolve
+suite: benchmark
+product:
+  - data api
+  - database
+  - auth
+topic:
+  - rls
+  - policies
```
mattrossman: Maybe will want to limit topic values to the set we've laid out in the [benchmark dimensions](https://app.notion.com/p/supabase/Supa-bench-condensed-3675004b775f8066824ce11e9fd06304?source=copy_link#3675004b775f8061ad05e804ec30d425) table, I realize rn this accepts arbitrary strings but eventually I'm gonna add validation so we can group by topics in the UI

evals/resolve-dataapi-001-empty-results/EVAL.ts
```diff
@@ -0,0 +1,237 @@
...
+  });
+
+  return {
+    name: "diagnosed RLS and added owner-scoped policies",
+    passed: verdict.passed,
+    judgeNotes: verdict.notes,
+  };
+}
+
+async function silenceExpectedError<T>(fn: () => PromiseLike<T>): Promise<T> {
```
mattrossman, line 227: Side note, maybe in a follow up we should pull this out into some shared framework logic, either as a shared function or something built into the framework so it silences scorer logs by default with option to enable for debugging.

mattrossman (approved):

evals/resolve-dataapi-001-empty-results/EVAL.ts
```diff
@@ -0,0 +1,237 @@
...
+  return {
+    name: "user B cannot insert a bookmark as user A",
+    passed: rows[0]?.count === 0,
+  };
+}
+
+async function checkRlsDiagnosisAndOwnerPolicies(
+  ctx: ToolEvalContext,
+): Promise<CheckResult> {
+  // AI-822: empty Data API results must be handled as an RLS issue, not a
```
mattrossman: A pattern I'm starting in https://github.com/supabase/evals/pull/16 is to note motivating citations in a `motivation: ` frontmatter field

mattrossman: In this case we may want to link to [this troubleshooting page](https://supabase.com/docs/guides/troubleshooting/why-is-my-select-returning-an-empty-data-array-and-i-have-data-in-the-table-xvOPgx) either in that field or in the associated issue somewhere

evals/resolve-dataapi-001-empty-results/PROMPT.md
Rodriguespn: Done ✅

evals/resolve-dataapi-001-empty-results/EVAL.ts
Rodriguespn: Agree

evals/resolve-dataapi-001-empty-results/EVAL.ts
Rodriguespn: Moved the conversation to https://github.com/supabase/evals/pull/22

PR #20, opened 2026-06-10
ci: run changed eval refreshes

mattrossman:
- Adds a `run-evals-changed` label path that runs only eval directories changed in the PR and preserves `run-evals` as the full benchmark refresh.
- Adds `export-results --merge` so partial exports replace refreshed `experiment`/`eval` rows without dropping unrelated web results.
- Verified with framework TypeScript compile and a local `export-results --merge` run against filtered raw results.

**References**
- https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
- https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions#contains


mattrossman:
Superseded by https://github.com/supabase/evals/pull/54

PR #22, opened 2026-06-11
refactor: silence expected scorer logs in the framework

Rodriguespn:
Follow-up to a review note on #17:

> Side note, maybe in a follow up we should pull this out into some shared framework logic, either as a shared function or something built into the framework so it silences scorer logs by default with option to enable for debugging.

Scorers routinely trigger supabase-js operations that are *expected* to fail (e.g. a cross-user write blocked by RLS), and `supabase-js` logs those failures via `console.error` even though the scorer handles them as values. 
Until now each eval worked around this with its own copy-pasted `silenceExpectedError` helper wrapped around the risky call.

Implements `runScorer` with a `--debug` option, passed as command line argument, that silences `console.error` for the duration of the scorer at a framework level. If `debug=true`, the supabase-js error is thrown.

**Notes**

- Scope is `console.error` only (the exact channel supabase-js uses), matching the old helper. Scorers report results via `CheckResult`/`notes`, not the console, so nothing meaningful is hidden — and `--debug` brings it all back.
- Relies on scorers running sequentially (they do, in both the runner loops and the smoke steps), since `console.error` is swapped globally.

### DIFF
--- a/apps/framework/harness/run-eval.ts
+++ b/apps/framework/harness/run-eval.ts
@@ -17,6 +17,7 @@ import {
   readRepeatedFlag,
   readSuiteFilters,
 } from "../lib/cli-args.js";
+import { runScorer } from "../lib/scorer.js";
 import { buildFileTools } from "./file-tools.js";
 import { viteBuild, vitestRun } from "./project-runner.js";
 import type {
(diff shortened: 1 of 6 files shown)


mattrossman (approved):
Thanks!

> Relies on scorers running sequentially

I'll make a note to revisit this silencing behavior in [AI-829](https://linear.app/supabase/issue/AI-829/parallel-eval-execution) as I'd like to support parallel execution down the road.

mattrossman:
Tested working by running this with & without the `--debug` flag, worked as intended!
```
pnpm eval -- --experiment openai-gpt-5.4-nano --eval resolve-security-001-rls-cross-user-leak --runs=1 --force --debug
```

PR #24, opened 2026-06-11
feat: add private Storage access eval (AI-810)

Rodriguespn:
See "Configure private Storage access" scenario from [Supa-bench (condensed)](https://app.notion.com/p/supabase/Supa-bench-condensed-3675004b775f8066824ce11e9fd06304?source=copy_link).

**Additions**
- Benchmark eval `build-storage-001-private-bucket-access` for configuring private per-user Storage access: a private `user-files` bucket, per-user upload and download RLS policies on `storage.objects`, and supabase-js code that shares a file through a temporary signed URL.
- platform-lite now provisions the storage schema for every project (next to the auth schema seeds): `storage.buckets`/`storage.objects` with column shapes verified against a fresh `supabase start`, RLS enabled with no policies, role grants, and `storage.foldername()` — plus tests covering the provisioned baseline and policy enforcement.

supabase-lite does not enforce RLS on its (experimental) Storage API, so the scorer verifies the policies through the Data API instead: test users created with `auth.signUp` and behavioral probes issued via `client.schema('storage')`, which lite serves with RLS enforced. The signed-URL step is verified from the transcript by the LLM judge. Worth revisiting with end-to-end Storage API checks (real upload/download/signed-URL expiry) once a runtime with enforcing storage exists.

**To run the eval**
```bash
pnpm eval -- --eval build-storage-001-private-bucket-access --experiment claude-sonnet-4.6 --force --runs 1
```

**What the eval checks**
- The `user-files` bucket exists and is private (`public = false`).
- RLS remains enabled on `storage.objects`.
- User A lists only their own files.
- User B cannot read user A's files.
- Anon reads no files.
- User A can upload into their own folder.
- User B cannot upload into user A's folder as user A.
- Agent kept the bucket private, wrote owner-scoped policies for the `authenticated` role, and provided sharing code that calls `createSignedUrl` with an expiry.

Closes AI-810


package.json
```diff
@@ -28,6 +28,7 @@
   "devDependencies": {
     "@supabase-evals/core": "workspace:*",
     "@types/common-tags": "^1.8.4",
+    "@types/node": "catalog:",
```
Rodriguespn, line 31: This was just to get the linting errors in my ide to disappear. Happy to remove if you think it's not necessary

mattrossman: You might need to run `Typescript: Select Typescript Version" and select the workspace version of TypeScript, I don't see linter errors when that one is selected but I do with the default of VS Code's version.

Image

Once that's selected, I'm seeing `node:*` imports click through to `node_modules/@types/node` as expected.

Although, it might be good for us to `catalog:` this dep regardless since I see we have a few different versions of it specified across the repo. I'd also expect the types version to align with the Node.js major version we have pinned in [`package.engines`](https://github.com/supabase/evals/blob/df510670b3e1c9bfeb03e84bf4db165acb7f2a73/package.json#L34) or [`.node-version`](https://github.com/supabase/evals/blob/df510670b3e1c9bfeb03e84bf4db165acb7f2a73/.node-version), i.e. 22

evals/build-storage-001-private-bucket-access/EVAL.ts
```diff
@@ -0,0 +1,278 @@
...
+import { stripIndent } from "common-tags";
+
+const BUCKET = "user-files";
+const USER_A = "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa";
+const USER_B = "bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb";
+const FILE_A1 = `${USER_A}/receipt-alpha.pdf`;
+const FILE_A2 = `${USER_A}/receipt-beta.pdf`;
+const FILE_B = `${USER_B}/receipt-gamma.pdf`;
+
+const asUser = (sub: string, body: string, finish: "COMMIT" | "ROLLBACK" = "COMMIT") => `
```
mattrossman: Something I tried in https://github.com/supabase/evals/pull/16 was shifting away from the user impersonation pattern, in favor of using the JS SDK to create users and then take action as them. I find it helpful for readability to not have the noise from transaction setup/teardown, and a little closer to how users would interact w/ the db.

E.g. `const client = ctx.getClient()` gives you an anon role client, or `client.signUp(...)` to get an authenticated client, then you can seed w/ SDK calls.

packages/platform-lite/src/project/storage-schema.ts
```diff
@@ -0,0 +1,61 @@
...
+-- policies (how a real project ships, and the deny-all starting state),
+-- grants so the API roles reach policy evaluation instead of failing on
+-- permissions, and storage.foldername() which path-scoped policies use.
+-- Production extras this eval never touches (other storage tables, helper
+-- functions, performance indexes, triggers) are omitted. The platform-lite
+-- runtime provisions no storage schema of its own, so this seed is what
+-- creates it.
+--
+
+CREATE SCHEMA storage;
```
mattrossman, line 10: Could this become part of the shared `platform-lite` setup? This mock of the storage schema feels broadly useful for the platform beyond this scenario.

We have some similar seeds for `auth` schema in:
https://github.com/supabase/evals/blob/df510670b3e1c9bfeb03e84bf4db165acb7f2a73/packages/platform-lite/src/project/ProjectInstance.ts

package.json
Rodriguespn: It's one of those "causality does not mean causation" situations, where I changed the typescript version to point to `node_modules` instead of using the global version after installing `@types/node`. Anyway, I `catalog:` the different node versions to v`^22.0.0` across the whole framework. Happy to move this to a separate PR if you think it's worth it

packages/platform-lite/src/project/storage-schema.ts
Rodriguespn: Good idea. I deliberately staid away from adding the storage schema at the framework level to reduce the entropy but I recognize that it's valuable to have the mocked storage schema at the framework level. LMK what you think
https://github.com/supabase/evals/pull/24/changes#diff-46ae449bbb83fb19a6cdb20396ac1c6bc25fcc1a226df19217c5ba6b85d5c97c

evals/build-storage-001-private-bucket-access/EVAL.ts
Rodriguespn: I wasn't sure if RLS would be enforced for the storage schema through the Data api because of supabase-lite but tested locally and it does. So now the pattern is the same as in #16. Thanks for the suggestion!

mattrossman (approved):

PR #25, opened 2026-06-11
feat: Supabase CLI eval backend in a Docker sandbox

Rodriguespn:
## Supabase CLI eval backend (local-stack Docker sandbox)

Lets eval agents drive the **real Supabase CLI** against a real local Supabase stack by running inside a Docker sandbox — the developer-machine half of Supabase. platform-lite continues to mock the remote/hosted platform, which this PR does not touch.

### How evals run
Two runtimes, chosen per eval:
- **Tools** — the agent works against platform-lite's MCP/tool surface, no sandbox. Used when an eval has neither a `local/` directory nor `interface: cli`.
- **Local-stack** — the agent works inside a Docker sandbox with the real Supabase CLI (`supabase init/start/db/...`), a `bash` tool, and file tools; `supabase start` spawns sibling service containers through the mounted Docker socket. Used when an eval ships a `local/` workspace **or** declares `interface: cli` (the latter covers bootstrap scenarios that start empty).

`interface` (`mcp` | `cli`) is a benchmark/KPI dimension, **not** the runtime switch — `local/` presence and `interface: cli` decide whether a sandbox boots.

### What's added
- **`@supabase-evals/sandbox`** — the Docker sandbox (docker-CLI shell-outs, no native deps) plus Supabase CLI setup: pinned CLI image, iptables DNAT so sibling-container ports are reachable on the sandbox's `127.0.0.1` (the CLI health-checks loopback), and teardown scoped to eval-created resources so a developer's own stacks are untouched. A single `docker cp` primitive backs workspace seeding (`copyHostDir`), export (`copyToHost`), and file writes.
- **`localStack: localStackRuntime()`** experiment field, declared like MCP servers and skills; experiments without it skip local-stack evals.
- **`LocalStackScorer`** context — runs commands, file reads, and SQL inside the same sandbox the agent used, plus `getClient()` against the running stack.
- **Host-side scoring for build/test evals** — after the agent finishes, the workspace is exported to the host (`docker cp`) and the repo-root `vite`/`vitest` run against it, so that toolchain needn't exist in the sandbox. This replaces the old host-only "project mode": evals with a `local/` workspace now run in the sandbox and are scored this way.
- **`services:`** frontmatter — which local-stack services boot; `supabase start -x` excludes the rest (including when the agent runs `start`). `services: []` = database only; omit for the full stack.

### Evals
- **`build-cli-001-bootstrap-app`** (benchmark) — bootstrap a new app: init a project, create a `todos` table via a migration, authenticated-only read with RLS, seed rows, verify through the local REST API.
- Migrated **`build-frontend-001-todos-app`** and **`deploy-database-001-prometheus-metrics`** from the removed project mode onto the local-stack runtime.

### Tests & CI
- 15 sandbox unit tests (part of `pnpm check`) + a docker-gated integration test (sandbox boot, pinned CLI install, headless `supabase init`, DNAT reachability, binary-safe `docker cp` seeding, teardown): `pnpm --filter @supabase-evals/sandbox test:docker`.
- CI: the [eval-refresh workflow](https://github.com/supabase/evals/actions/runs/27618667317) runs the CLI evals end-to-end on a GitHub-hosted `ubuntu-latest` runner (via the `run-evals` label) and refreshes results — Docker, `NET_ADMIN`, and `supabase start` all work on hosted runners.

### Run it
```bash
pnpm eval -- --eval build-cli-001-bootstrap-app --experiment claude-sonnet-4.6 --force --runs 1
```
Requires a running Docker daemon and Supabase's default host ports (54321–54329) free.

### Out of scope
Linking the CLI to a hosted project (`supabase link`, `functions deploy`, `secrets set`, `db push --linked` against platform-lite) is **not** in this PR — tracked separately (AI-838, with the edge-function-secrets benchmark in AI-815). The local/remote naming anticipates that work.

### References
- AI-793 — Supabase CLI eval backend · AI-837 — local-stack runtime + CLI tool surface · AI-809 — bootstrap benchmark
- https://github.com/supabase/agent-eval — origin of the sandbox + CLI approach

Closes AI-837, AI-809


evals/build-cli-001-bootstrap-app/PROMPT.md
```diff
@@ -0,0 +1,30 @@
...
+services:
+  - gotrue
+  - kong
+  - postgrest
+motivation: AI-809, https://supabase.com/docs/guides/getting-started/tutorials/with-nextjs
+---
+
+# Todos App Needs Its Backend
+
+We're kicking off a todos app and I want the Supabase side ready for the team
```
Rodriguespn, line 20: Should we have the agent create the backend and link it to an already existing nextjs app that is missing supabase-js client (creating the app is NOT in the agent's duties. The frontend files would already be in the agent workspace) or let the agent just create the backend?

apps/framework/harness/run-eval.ts
```diff
@@ -291,6 +310,62 @@ async function runOne(
       continue;
     }
 
+    if (ev.mode === "cli") {
```
Rodriguespn: This function is starting to get very big and has repeated configurations for each `env.mode`. Worth considering a refactoring

packages/sandbox/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# Sandbox image for cli-interface evals: a developer machine with the
+# Supabase CLI installed. Built by ensureSupabaseSandboxImage() with
+# --build-arg CLI_VERSION pinned for benchmark comparability (installed from
+# the official release .deb; Homebrew's tap only carries the latest version).
+FROM node:22-slim
+
+RUN apt-get update -qq && apt-get install -y -qq --no-install-recommends \
+    ca-certificates curl git docker.io postgresql-client iptables iproute2 \
```
mattrossman, line 8: On my end, the `test` script works, but `pnpm --filter @supabase-evals/sandbox test:docker` fails with:

```
 FAIL  test/docker.test.ts > local-stack sandbox (docker) > boots a sandbox, starts a minimal local stack, and reaches it
AssertionError: expected '' to be '00ff00fe' // Object.is equality

- Expected
+ Received

- 00ff00fe

 ❯ test/docker.test.ts:68:37
     66|           await sandbox.copyHostDir(seed);
     67|           const hex = await sandbox.runShell("xxd -p bin.dat");
     68|           expect(hex.stdout.trim()).toBe("00ff00fe");
       |                                     ^
     69|           expect(await sandbox.fileExists(".git/HEAD")).toBe(false);
     70|         } finally {
```

Looks like adding `xxd` to the list of packages installed in the Dockerfile fixes it:

```suggestion
    ca-certificates curl git docker.io postgresql-client iptables iproute2 xxd \
```

Note I also had to `docker rmi supabase-evals-sandbox:2.67.1` after that change to get it to rebuild the image.

packages/core/src/eval-metadata.ts
```diff
@@ -45,6 +76,9 @@ export const evalMetadataSchema = z.object({
   product: z.array(evalProductSchema).min(1),
   topic: z.array(z.string().min(1)).min(1),
   suite: evalSuiteSchema.default("regression"),
+  environment: evalEnvironmentSchema.optional(),
+  services: z.array(z.string().min(1)).min(1).optional(),
```
mattrossman: What if you only need the default ["db"](https://github.com/supabase/cli/blob/75827499ca2940aa476b132f4389a6deb6b890ef/apps/cli-go/internal/utils/config.go#L104) service, no other services? Rn this forces you to specify at least one additional service. Maybe we should allow an empty list here to mean that you only need the database, no additional services.

evals/build-cli-001-bootstrap-app/PROMPT.md
mattrossman: Ideally there's some frontend component here, since this benchmark is intended to test the `sdk` / `data_api` topics, though we have other scenarios covering those topics so not essential if it complicates things much. I could see the scenario getting pretty slow if it's trying to populate `node_modules` and set everything up from scratch.

mattrossman:
I'm adding the `run-evals` label because aside from the `test:docker` run you showed, I want to understand whether this can actually run the CLI evals in CI and refresh results normally.

mattrossman:
Summarizing other feedback from Slack, I'd like to:
- keep `interfaces:` metadata and use that to determine whether Supabase CLI is available
- use `local/` directory existence to determine whether to use sandbox environment
  - OR if simpler use `interfaces: cli` to decide both
- replace the legacy "project mode" with the new sandbox environment

packages/sandbox/Dockerfile
Rodriguespn: Replace `xxd` with `od` package as `od` already comes with `node:22` docker image

Rodriguespn:
> I'm adding the `run-evals` label because aside from the `test:docker` run you showed, I want to understand whether this can actually run the CLI evals in CI and refresh results normally.

The [eval-refresh workflow](https://github.com/supabase/evals/actions/runs/27618667317) passed

packages/core/src/eval-metadata.ts
Rodriguespn: Fixed. Now if `services = []`, all services run

evals/build-cli-001-bootstrap-app/PROMPT.md
Rodriguespn: We don't necessarily need to install the frontend `node_modules` for this scenario, unless we want to test if the frontend code actually builds, but there is a high chance that the agent will run any typechecker/linter and get to the conclusion that it needs the `node_moduels` locally. So, I decided not to include the frontend for this scenario to simplify things

README.md
```diff
@@ -53,7 +53,7 @@ pnpm web
 2. Add `PROMPT.md` with frontmatter metadata and the task the agent sees.
 3. Add `EVAL.ts` with the scorer.
 4. Add `remote/` data if the scenario needs hosted-project state (database, logs, functions).
-5. Add `local/` files if the agent starts from an existing workspace (project evals: the app it edits).
+5. Add `local/` files if the agent starts from an existing workspace (project evals: the app it edits; CLI evals: the repo).
```
mattrossman: Maybe outdated "project evals" reference since IIUC this is now merged w/ the concept of CLI evals

evals/deploy-database-001-prometheus-metrics/EVAL.ts
```diff
@@ -36,7 +36,7 @@ const scorer: ProjectScorer = async (ctx) => {
     judge({
       input,
       rubric: stripIndent`
-        Pass if prometheus.yml adds a deployable Supabase Metrics API scrape for ${ctx.ref}.supabase.co or ${ctx.ref}.supabase.red.
```
mattrossman, line 39: Previously in "project mode" evals, agents could talk to platform-lite via MCP.

Do we no longer do that? I wonder how that impacts this test, as before we were wiring up to a "remote" project but it looks like we no longer have that ref in context. I imagine that also means agents don't have docs access.

apps/framework/harness/run-eval.ts
```diff
@@ -226,51 +232,64 @@ async function runOne(
...
         const run = await exp.agent.run({
-          systemPrompt,
+          systemPrompt: buildSystemPrompt(
+            "local-stack",
+            skillContext,
+            session.promptAddendum,
+          ),
           userPrompt: prompt,
-          tools: fileTools,
-          mcpServers: session.mcpServers,
```
mattrossman, line 246: Around here is where I think we'd still want to expose MCP servers

mattrossman (approved):

apps/framework/harness/run-eval.ts
Rodriguespn: As discussed on Slack, addressing this on another PR. Will post here the link to that PR when available

evals/deploy-database-001-prometheus-metrics/EVAL.ts
Rodriguespn: Answered below

README.md
Rodriguespn: Fixed, thx for the callout

apps/framework/harness/run-eval.ts
Rodriguespn: https://github.com/supabase/evals/pull/39

PR #27, opened 2026-06-11
feat: add Realtime publication debugging eval (AI-819)

Rodriguespn:
See "Debug Realtime publication" scenario from [Supa-bench (condensed)](https://app.notion.com/p/supabase/Supa-bench-condensed-3675004b775f8066824ce11e9fd06304?source=copy_link#37c5004b775f804ebab9e23b92f6848b).

**Additions**
- Benchmark eval `investigate-realtime-001-subscribed-no-events` for debugging a Realtime subscription that reaches SUBSCRIBED but never receives INSERT events, tracing the cause to the table missing from the `supabase_realtime` publication and fixing it with a targeted `ALTER PUBLICATION ... ADD TABLE`.
- Seed with a working dispatch app: a courier locations feed whose table is in the publication (live updates work) and an `orders` table that was never added (the bug). The symptom — SUBSCRIBED with no events — is only observable client-side, so the prompt reports it the way a user would; the diagnosis and fix are entirely database-side, where the agent can work through the Management API.

Once a runtime with a real Realtime service exists, this eval can be upgraded to verify end-to-end delivery (subscribe, insert, assert the event arrives after the fix).

**To run the eval**
```bash
pnpm eval -- --eval investigate-realtime-001-subscribed-no-events --experiment claude-sonnet-4.6 --force --runs 1
```

**What the eval checks**
- The `orders` table was added to the `supabase_realtime` publication.
- `courier_locations` is still in the publication (the working feed survived the fix).
- The publication still publishes INSERT events.
- RLS remains enabled on `orders` and staff can still read orders through it.
- Agent diagnosed the silent subscription as missing publication membership (not RLS, client code, or networking) and fixed exactly that, without recreating the publication `FOR ALL TABLES` or recommending read replicas.

**References**
- https://linear.app/supabase/issue/REAL-577/avoid-retrying-to-reconnect-pg-changes-for-unrecoverable-errors

Closes AI-819

### DIFF
--- a/evals/investigate-realtime-001-subscribed-no-events/EVAL.ts
+++ b/evals/investigate-realtime-001-subscribed-no-events/EVAL.ts
@@ -13,7 +13,8 @@ const scorer: ToolScorer = async (ctx) => {
       await checkOrdersInPublication(ctx),
       await checkCourierFeedStillPublished(ctx),
       await checkPublicationStillPublishesInserts(ctx),
-      await checkOrdersRlsIntact(ctx),
+      await checkOrdersRlsEnabled(ctx),
+      await checkStaffCanReadOrders(ctx),
       await checkPublicationDiagnosis(ctx),
     ];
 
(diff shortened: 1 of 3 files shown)


evals/investigate-realtime-001-subscribed-no-events/EVAL.ts
```diff
@@ -0,0 +1,144 @@
...
+    `SELECT relrowsecurity FROM pg_class WHERE relname = 'orders';`,
+  );
+  if (rls[0]?.relrowsecurity !== true) {
+    return { name: "RLS still enabled on orders", passed: false };
+  }
+
+  try {
+    const { rows } = await ctx.query(stripIndent`
+      BEGIN;
+      SET LOCAL ROLE authenticated;
```
mattrossman: Similar to https://github.com/supabase/evals/pull/24#discussion_r3399568240, wondering if we can use JS client to run authenticated queries vs impersonation.

evals/investigate-realtime-001-subscribed-no-events/EVAL.ts
```diff
@@ -0,0 +1,144 @@
...
+      SET LOCAL request.jwt.claim.role = 'authenticated';
+      SELECT count(*)::int AS count FROM orders;
+      COMMIT;
+    `);
+
+    return {
+      name: "RLS still enabled on orders",
```
mattrossman: Wonder if this should be a separate / renamed check, cause this is more checking whether staff can see orders? The earlier `rls[0]?.relrowsecurity` check seems more like the "RLS still enabled" part.

Side note, maybe we can tighten this alongside the JS client comment above by comparing what authenticated client sees against what superuser sees instead of hardcoding 2 in case agent decides to insert test values during the run.

mattrossman (approved):

evals/investigate-realtime-001-subscribed-no-events/EVAL.ts
Rodriguespn: Split into two checks as suggested:
1. Is RLS enabled
2. Staff can still see the orders through the Data api

Thx for the suggestion

PR #29, opened 2026-06-12
feat: add deleted user session revocation eval (AI-820)

Rodriguespn:
See "Revoke deleted user sessions" scenario from [Supa-bench (condensed)](https://app.notion.com/p/supabase/Supa-bench-condensed-3675004b775f8066824ce11e9fd06304?source=copy_link#37d5004b775f80d4921de117225ecdae).

**Additions**
- Benchmark eval `investigate-auth-001-deleted-user-access` for debugging why a "deleted" account keeps working: the app's `delete_account` function only soft-deletes the profile row and never touches the auth user or their sessions. The agent must diagnose the flow, implement real revocation, explain the remaining access-token validity window, and clarify publishable vs secret API keys and their RLS implications.
- Seed with the broken flow: `profiles`/`notes` with owner-scoped RLS and a `SECURITY DEFINER` `delete_account()` that flips `profiles.deleted` and nothing else.

The premise (stale JWTs keep working after deletion) is backed by the [Max-Age cookies troubleshooting page](https://supabase.com/docs/guides/troubleshooting/should-i-set-a-shorter-max-age-parameter-on-the-cookies-8sbF4V): 

- local JWT validation only checks signature and expiry, so server-side session end is invisible to it. Note that supabase-lite is stricter than production here — its Data API checks the user/session per request, so the access-token window itself cannot be reproduced in the runtime; it is scored from the explanation by the LLM judge. 
- The revocation effects (sessions, refresh tokens, sign-in) are verified behaviorally. The scorer runs the flow exactly as the app would: the signed-in victim calls `client.rpc("delete_account")` through supabase-js. This needed supabase-community/lite#243 (route `/rest/v1/rpc/*` to the postgrest layer), which is merged and shipped in `@supabase/lite` v0.4.1-next.2 (#31). The whole scorer uses the JS SDK pattern from #16.

**To run the eval**
```bash
pnpm eval -- --eval investigate-auth-001-deleted-user-access --experiment claude-sonnet-4.6 --force --runs 1
```

**What the eval checks**
- The victim's session works before the delete-account flow runs.
- The `delete_account` flow runs for the victim.
- The flow revokes the victim's sessions.
- The victim's refresh token is rejected afterwards.
- The victim cannot sign back in.
- Other users keep their sessions and access.
- Agent diagnosed the soft-delete-only flow, implemented real revocation, explained the stateless access-token window (and how to shrink it), and correctly placed publishable vs secret keys with their RLS implications.

Closes AI-820

### DIFF
--- a/evals/investigate-auth-001-deleted-user-access/PROMPT.md
+++ b/evals/investigate-auth-001-deleted-user-access/PROMPT.md
@@ -9,7 +9,7 @@ topic:
 motivation: AI-820, AI-422, https://supabase.com/docs/guides/troubleshooting/should-i-set-a-shorter-max-age-parameter-on-the-cookies-8sbF4V
 ---
 
-# Deleted User Still Has Access
+# Revoke deleted user sessions
 
 Last week support removed a user through our app's delete-account flow — the
 app calls the `delete_account` function over RPC as the signed-in user. This
--- a/evals/investigate-auth-001-deleted-user-access/PROMPT.md
+++ b/evals/investigate-auth-001-deleted-user-access/PROMPT.md
@@ -9,8 +9,6 @@ topic:
 motivation: AI-820, AI-422, https://supabase.com/docs/guides/troubleshooting/should-i-set-a-shorter-max-age-parameter-on-the-cookies-8sbF4V
 ---
 
-# Revoke deleted user sessions
-
 Last week support removed a user through our app's delete-account flow — the
 app calls the `delete_account` function over RPC as the signed-in user. This
 morning that same person was back: still signed in, reading and saving their
--- a/evals/investigate-auth-001-deleted-user-access/PROMPT.md
+++ b/evals/investigate-auth-001-deleted-user-access/PROMPT.md
@@ -1,0 +1,26 @@
+---
+stage: investigate
+suite: benchmark
+product:
+  - auth
+topic:
+  - security
+  - sdk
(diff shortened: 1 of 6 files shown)


evals/investigate-auth-001-deleted-user-access/PROMPT.md
```diff
@@ -0,0 +1,26 @@
...
+suite: benchmark
+product:
+  - auth
+topic:
+  - security
+  - sdk
+motivation: AI-820, AI-422, https://supabase.com/docs/guides/troubleshooting/should-i-set-a-shorter-max-age-parameter-on-the-cookies-8sbF4V
+---
+
+# Revoke deleted user sessions
```
mattrossman: Wondering if this header gives the agent a head start on this check:
https://github.com/supabase/evals/pull/29/changes#diff-f5972c715f67421ef8c16ced9eea206a50c63e4fe6197a5a58ca8a06da9ded6eR213

As I'm adding new evals, I'm omitting these header blocks b/c I don't want to tip off what we're testing in the scenario, and I'm trying to model these as just the text a user would be typing into a prompt.

Rodriguespn: Yup, the heading is passed to the agent in `parseEvalMarkdown` and it's a clue for the agent. I've removed the heading from this PR but I'm opening another one to remove the toher headings from the prompts as we have a bunch of them

Rodriguespn: https://github.com/supabase/evals/pull/32 🚀

PR #33, opened 2026-06-15
feat: add slow query CPU spike eval

mattrossman:
Repurposes the log-based slow-query eval `resolve-performance-001-slow-query-index` into a benchmark `resolve-performance-001-slow-query-cpu-spike` for the CPU spike scenario. This was based on multiple motivating feedback tickets around difficulty understanding how to diagnose CPU spikes, and our [troubleshooting guide](https://supabase.com/docs/guides/troubleshooting/high-cpu-usage) here.

I explored if we could leverage the [index_advisors](https://supabase.com/docs/guides/database/extensions/index_advisor) workflow (or advisors as a whole) but that'd involve more siginficant changes to support in supalite (see [STATUS.md](https://github.com/supabase-community/lite/blob/6a33418ccd89a210c2bb52dbfec2a43126fab234/STATUS.md#db-group)). Instead, this seeds the remote project with `pg_stat_statements-style` query stats and scores both the intended diagnostic path and final database state. I'm keeping the existing `logs.jsonl` seed around for good measure though it's not really needed for the solution.

**What the scenario is**

This scenario seeds a Supabase project with an `events` table, slow-query Postgres logs, and `pg_stat_statements`-style query performance data showing repeated expensive recent-events lookups for a user. The agent is asked to investigate the CPU spike / slow app behavior, identify the offending query, confirm the poor plan with `EXPLAIN`, and apply the missing database index fix.

**What the eval checks**
- The agent inspected `pg_stat_statements` / query performance data.
- The agent ran `EXPLAIN` on the expensive `events` query.
- The database has an index on `events` covering `(user_id, created_at)`.
- The target recent-events query uses an index in its query plan.
- The target recent-events query does not use a sequential scan on `events`.
- Inserts into `events` still work after the schema change.

**Results**

Nano passed, Mini failed as it didn't check `pg_stat_statements`

Closes AI-824


apps/web/src/data/eval-results.json
```diff
@@ -64,19 +123,65 @@
...
+    "stage": "investigate",
+    "product": [
+      "realtime",
+      "database"
+    ],
+    "topic": [
+      "sdk"
+    ],
+    "suite": "benchmark",
+    "passed": false,
```
Rodriguespn, line 151: nit: In the PR description is`Both agents tested (5.4 mini/nano) pass` but gpt-5.4-mini failed in this experiment

mattrossman: That failure is from a different eval, the `investigate-realtime-001-subscribed-no-events` one!

The `resolve-performance-001-slow-query-cpu-spike` one is passing for both, see Preview:

CleanShot 2026-06-15 at 12 17 00@2x

Rodriguespn:
I wonder if this scenario is significant enough to prove that the agent correctly *diagnosed* the CPU spike. It fixes the spike by adding an index, but checking mini's [transcript](https://github.com/supabase/evals/actions/runs/27558240880/artifacts/7643436937) we can see it never queried `pg_stat_statements`, it called `get_advisors` (which returns `{"lints":[]}` here), then `get_logs`, and went straight to `apply_migration`. It passes every outcome check (index created, plan uses it, no seq scan, inserts work) but fails inspected `pg_stat_statements`.

In this case the logs are enough for the agent to land on the right query: each `q*` line spells out the full SQL (`SELECT id, kind, payload, created_at FROM events WHERE user_id = $1 ORDER BY created_at DESC LIMIT 50`) at 840–2140 ms while the other queries log at 12–18 ms, so it's obviously *the* slow one. But the logs only carry **per-execution latency** — they're missing the **call frequency / cumulative time** (`calls`, `total_exec_time`) that lives in `pg_stat_statements`, which is really the essence of a CPU diagnosis. A query is a CPU problem because of `mean_exec_time × calls`, not because any single execution was slow.

Imagine the real CPU hog is a query that runs in ~25 ms each time but is called 800k times (total ≈ 20,000 s of CPU), alongside a genuinely slow but rare report query that takes 3 s but runs 10 times a day (total ≈ 30 s). The slow-query logs would surface the 3 s report (and likely never show the 25 ms query at all, since it's under the `log_min_duration_statement` threshold), pointing the agent at the *wrong* query. Only `pg_stat_statements`, sorted by `total_exec_time`, reveals that the high-frequency 25 ms query is what's actually pinning the CPU. 
I think it's the skill this eval really wants to measure.

Curious whether you intended the logs to be a valid alternate path here, or whether the goal is specifically to exercise the `pg_stat_statements` workflow — happy to push either change.

mattrossman:
@Rodriguespn Yes, the goal here is to use the  `pg_stat_statements` workflow. I kept the logs seed around since they were in the existing eval, and I figured if agent decided to check logs at least it should have something semi realistic to look at.

There's a check `"inspected pg_stat_statements for query performance"` for this, and I realize that your earlier comment is actually correct that mini failed (because of that check). I think the Vercel preview comment didn't update (it links to the [branch preview URL](https://evals-git-feat-ai-824-slow-query-cpu-spike-supabase.vercel.app/) which is still showing mini passing) but when I visit the [commit preview URL](https://evals-dhn4na6ve-supabase.vercel.app/) from further down on the page, I see the mini failure as expected.

CleanShot 2026-06-15 at 13 59 15@2x
CleanShot 2026-06-15 at 14 04 29@2x

Updating the PR body to reflect that result!

Rodriguespn (approved):
LGTM

PR #38, opened 2026-06-16
feat: add Edge Function secrets deploy eval (AI-815)

Rodriguespn:
> Stacked on #39 (`search_docs` MCP for the local-stack runtime) - base is that branch, so this diff is only the AI-815 changes. Retarget to `main` once #39 lands.

See ["Move a leaked API key into an Edge Function secret" scenario](https://app.notion.com/p/supabase/Supa-bench-condensed-3675004b775f8066824ce11e9fd06304?source=copy_link#3815004b775f80228711ccaefafde669).

**Additions**
- Benchmark eval `deploy-functions-001-edge-function-secrets` (CLI interface, hosted project): a weather widget leaks a WeatherAPI key from the browser; the agent moves it behind a `weather` Edge Function that reads `WEATHER_API_KEY` from a Function secret, deploys it live, and keeps the key out of the repo. The work is entirely project-side, driven through the Supabase CLI (`supabase secrets set`, `supabase functions deploy`).
- **platform-lite Management API**, so the flow is verifiable end to end:
  - `secrets` — `GET`/`POST`/`DELETE /v1/projects/{ref}/secrets` over a per-project store. List returns the SHA-256 digest (never plaintext) and rejects the reserved `SUPABASE_` prefix, mirroring the real API; plaintext is kept server-side and exposed to the in-process function runtime as env, as the hosted Edge Runtime does.
  - `functions` — create `POST /v1/projects/{ref}/functions` and redeploy `PATCH /v1/projects/{ref}/functions/{slug}` (eszip body + metadata in query params). These are exactly the calls `supabase functions deploy` makes, **verified against the live Management API**: a new slug is created with `POST`, an existing one updated with `PATCH`.
- **Linked without running `supabase link`.** The runtime seeds the post-link end state directly — `supabase/.temp/project-ref` (the file `link` writes), a CLI profile (`api_url`), and `SUPABASE_ACCESS_TOKEN` — so `functions deploy` / `secrets set` target platform-lite. Gated by the `hostedProject` frontmatter flag. (The rest of `link`'s state is only needed for Postgres-wire workflows like `db push`, tracked separately in AI-843.)
- **MCP for the agent.** The local-stack runtime exposes a docs-only Supabase MCP server (`search_docs`) — the sandbox has no web tools. Other feature groups are withheld on purpose so the agent can't bypass the CLI workflow under test by deploying through MCP.

**To run the eval**
```bash
pnpm eval -- --eval deploy-functions-001-edge-function-secrets --experiment claude-sonnet-4.6 --force --runs 1
```

**What the eval checks**

The scorer reads ground truth from platform-lite via the management client plus the workspace files:
- `WEATHER_API_KEY` is set as a Function secret on the project.
- The `weather` function is deployed and `ACTIVE`.
- The deployed function reads `WEATHER_API_KEY` from the environment (`Deno.env.get(...)`), not from request input or a config file.
- The key is not committed: the source does not hardcode it and no tracked file assigns it a value.

Closes AI-815


evals/deploy-functions-001-edge-function-secrets/EVAL.ts
```diff
@@ -0,0 +1,153 @@
...
+      ],
+    };
+  }
+};
+
+export default scorer;
+
+// The secret was set on the hosted project. Read it from the management API,
+// which never returns plaintext — its presence by name is the assertion.
+async function checkSecretSet(ctx: LocalStackEvalContext): Promise<CheckResult> {
```
mattrossman, line 70: What is the expected path for the agent to pass this check? We don't include a secret in the seed or prompt, so how would it know what value to `set` the secret to?

Maybe we should seed this scenario with an `.env` file so the agent can use `supabase secrets set --env-file` to set a known secret value and avoid leaking it into context (and maybe we check that it doesn't read this secret in context)

evals/deploy-functions-001-edge-function-secrets/EVAL.ts
```diff
@@ -0,0 +1,153 @@
...
+    passed: readsEnv,
+    notes: readsEnv ? undefined : `no Deno.env.get("${SECRET_NAME}") found in the function source`,
+  };
+}
+
+// The key must not be committed: the function source must not hardcode it, and
+// any env file holding it must be gitignored (the CLI's `supabase init` already
+// gitignores supabase/.env). We can't know the value the agent chose, so we
+// flag assignment-to-a-literal patterns rather than the env-read reference.
+async function checkSecretNotInRepo(ctx: LocalStackEvalContext): Promise<CheckResult> {
```
mattrossman, line 143: It looks like this is checking whether the variable name e.g. `WEATHER_API_KEY` appears in the function source, but what if the agent wrote the secret value without assigning it to a variable e.g. `getWeather("secret_abc123")`, or used a different variable name?

It's hard to be 100% correct here as there's any number of ways of leaking secrets into source code but I figure we should at least check if the secret text appears in the source.

mattrossman: Seeding with a fixed secret value could address this one too https://github.com/supabase/evals/pull/38/changes#r3430467400

evals/deploy-functions-001-edge-function-secrets/EVAL.ts
Rodriguespn: Adding a seeded `.env` with a mock api key is a good idea, but we might get flagged by Github for having a `.env` file in the repo. If this happens, I suggest creating a hook that allow us to edit files inside the container to create the `.env` after copying the workspace to the container and before the agent runs

evals/deploy-functions-001-edge-function-secrets/EVAL.ts
Rodriguespn: What if instead of a check, we delegate this to a judge to check if the agent correctly imported and read the env variable?

mattrossman: That works for me!

mattrossman (approved):

PR #39, opened 2026-06-16
feat(eval): expose `search_docs` MCP to the local-stack runtime

Rodriguespn:
Gives local-stack (sandbox/CLI) evals access to the Supabase MCP `search_docs` tool. `localStackRuntime()` now starts a docs-only Supabase MCP server (`--features docs`) host-side and the runner merges it with the sandbox tools, so the agent can search docs even when the stack isn't running (e.g. the prometheus eval).

**Test:**
```
pnpm eval --experiment=openai-gpt-5.4-mini --eval=deploy-database-001-prometheus-metrics --force
```

### DIFF
--- a/packages/sandbox/src/local-stack-runtime.ts
+++ b/packages/sandbox/src/local-stack-runtime.ts
@@ -1,9 +1,11 @@
 import { posix } from "node:path";
 import { jsonSchema, tool, type ToolSet } from "ai";
 import { createClient } from "@supabase/supabase-js";
-import type {
-  LocalStackRuntime,
-  LocalStackScoringContext,
+import {
+  docsMcpServer,
+  type LocalStackRuntime,
+  type LocalStackScoringContext,
+  type McpServerConfig,
 } from "@supabase-evals/core";
 import { DockerSandbox } from "./docker-sandbox.js";
 import {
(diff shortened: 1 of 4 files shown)


packages/sandbox/src/local-stack-runtime.ts
```diff
@@ -33,11 +35,19 @@ const MAX_TOOL_OUTPUT_CHARS = 16_000;
...
+   * `search_docs` (the sandbox has no web tools). Pass `{}` to disable, or add
+   * more servers. These run host-side; they do not connect to the sandbox.
+   */
+  mcpServers?: Record<string, McpServerConfig>;
 }
 
 export function localStackRuntime(
   options: LocalStackRuntimeOptions = {},
 ): LocalStackRuntime {
+  const mcpServers = options.mcpServers ?? { "supabase-docs": docsMcpServer() };
```
Rodriguespn: Since this will be replaced by supabase MCP server, added here as a default instead of adding it to every experiment file

mattrossman: Wondering if we should omit "docs" from the other MCP server in the meantime to avoid duping that tool

Rodriguespn: New approach, make `--api-key` an optional argument and initialize the MCP server with docs only. The `search_docs` tool only depends on the GraphQL content api, so we don't have to worry about which platform the MCP is pointing to, as the only tool used doesn't depend on it 

Working on a [PR](https://github.com/supabase/evals/pull/38) to have platform-lite running alongside the local stack

mattrossman (approved):
Thanks!

PR #41, opened 2026-06-17
feat(benchmark): add declarative schema eval

mattrossman:
Adds a benchmark `build-cli-002-declarative-schema` eval for the declarative schema workflow. Motivated by Saxon's [feedback](https://supabase.slack.com/archives/C051L8U2EJF/p1779406337.608509?thread_ts=1779398780.194929&cid=C051L8U2EJF) that we should probe how well agents handle Supabase's composable workflows w/ declarative schemas being a primary example.

**What the scenario is**

Seeds a Supabase project with a `products` table already defined via `supabase/schemas/products.sql` + a corresponding migration already applied. The agent is asked to add a `description` column, the hope being that the agent recognizes the existing declarative schema pattern and follows it.

**What the eval checks**
- `supabase db diff` was invoked (agent followed the declarative workflow)
- The schema file was updated to include `description`
- A new migration was generated
- The `description` column is live in the database

**Results**

Both models tested failed to use the declarative schema workflow, instead opting to manually write a new migration. This might be due to the lack of docs support, so worth re-running after https://github.com/supabase/evals/pull/39 is merged.

CleanShot 2026-06-16 at 22 13 06@2x

CleanShot 2026-06-16 at 22 13 10@2x

**References**
- https://supabase.com/docs/guides/local-development/declarative-database-schemas

Closes AI-814


Rodriguespn:
Rebased this branch with origin/main to include the commit in main with the supabase MCP enabled for local stack runtime and trigger a new evals run

Rodriguespn:
Looking at the [eval results](https://github.com/supabase/evals/actions/runs/27692555648), neither of the models called `search_docs`, even though the tool is now available. This is a pattern that we've identified in agents where they are too lazy to consult external sources unless instructed to do so. Wonder if the agents loaded our skill the results would improve

PR #43, opened 2026-06-18
feat(eval): db push migration-history-mismatch eval + Postgres-wire backend

Rodriguespn:
See "Recover migration history mismatch" scenario from [Supa-bench (condensed)](https://app.notion.com/p/supabase/Supa-bench-condensed-3675004b775f8066824ce11e9fd06304?source=copy_link#37b5004b775f800c92abe5bdae006b0b).

**Additions**
- Benchmark eval `resolve-database-001-migration-history-mismatch` for recovering a remote migration-history mismatch — versions exist on the hosted database but are missing locally, so `supabase db push` fails with "The remote database's migration history does not match local files." The agent must reconcile and ship a pending migration **without resetting production data**.
- Seed with a hosted project whose history is ahead of local in a realistic way: a `bio` column was applied straight on the hosted DB and never committed, while the local workspace holds the original `create_profiles` and a pending `add_avatar_url`. The mismatch is real over the Postgres wire, and 25 production profiles must survive the fix.

### Remote database connection

platform-lite now lets the agent's Supabase CLI reach the mocked hosted project's database over the Postgres wire, not just the HTTP Management API — so linked DB workflows (`db push`, `db pull`, `migration repair`) work end-to-end (AI-843).

- **Postgres-wire surface:** built on `@electric-sql/pglite-socket`, with a small shim that answers the `SSLRequest`/`GSSENCRequest` probes libpq/pgx send (pglite-socket only frames the v3 StartupMessage), and tenant routing (`postgres.<ref>`) so one listener serves every project.
- **Two-surface platform-lite:** `PlatformHandle` exposes `listen()` (HTTP gateway) and `listenPg()` (the pooler) as peers over one PGlite, mirroring a real project. The sandbox link seam seeds `supabase/.temp/pooler-url` + version files so the agent's CLI dials it transparently.
- **Per-connection prepared-statement reset:** PGlite is a single shared session, but the CLI's pgx driver names prepared statements with a process-local counter (`lrupsc_1_0`), so each fresh `supabase` invocation re-`Parse`s the same name and collided with the previous one's leftover (`prepared statement already exists`, SQLSTATE 42P05) — which broke every `db push`/`migration repair` after the first. The pooler now `DEALLOCATE ALL`s on the 0→1 client transition, emulating the per-connection session a real backend gives (the CLI uses session-mode port 5432 precisely because it relies on prepared statements). Covered by an extended-protocol regression test.
- **`ctx.hostedQuery`:** the in-process read path — runs SQL directly against the hosted project's PGlite (the hosted counterpart to the local-stack `query`), used for ground-truth scoring without a Management API round-trip.

This is verified end-to-end: an agent reconciles the history and runs a real `supabase db push` over the wire (`Applying migration … / Finished supabase db push`, no `42P05`), passing all checks. The wire transport, SSL negotiation, tenant routing, per-connection reset, and the scorer's ground-truth reads are also covered by unit tests.

**To run the eval**
```bash
pnpm eval -- --eval resolve-database-001-migration-history-mismatch --experiment claude-sonnet-4.6 --force --runs 1
```

**What the eval checks** (ground truth read via `ctx.hostedQuery`, never the CLI under test)
- The `avatar_url` column is applied on the hosted `profiles` table.
- The pending migration `20240220` is recorded in the remote history (pushed, not applied as ad-hoc SQL).
- The remote history is reconciled with local files (no remote-only orphan left).
- The local migrations are a valid reconciled sequence — accepts **either** valid solution: the orphan recovered into its `20240115` slot, **or** a `db pull`-style bio captured under any timestamp before the avatar. Both require `create_profiles` first, `add_avatar_url` last, exactly one bio reconciliation strictly between, strictly ascending timestamps, all applied on the remote. Rejects discarding the orphan, a bio after the avatar, reordering, a renamed seeded file, or an un-applied reconciliation.
- Production profile data is intact (all 25 profiles with their `bio`) — the remote was reconciled, not reset/wiped.
- **The fix went through the CLI workflow, not around it** — a judge over the agent's tool calls fails runs that reach the end state via the Management API, direct SQL, or other non-CLI paths instead of `supabase db push` / `migration repair`. Read-only inspection by any means stays allowed. Without this, an end-state pass can't distinguish "used the intended workflow" from "found a workaround."

Closes AI-823, AI-843

### DIFF
--- a/apps/framework/harness/run-eval.ts
+++ b/apps/framework/harness/run-eval.ts
@@ -380,10 +380,14 @@ async function runOne(
       }
       // When the eval links to a hosted project, boot a platform-lite backend
       // (bound to 0.0.0.0 so the sandbox reaches it via host.docker.internal)
-      // and hand the CLI-valid ref/token to the session.
+      // and hand the CLI-valid ref/token to the session. Seed it from the eval's
+      // `remote/` dir (project.sql / logs.jsonl / functions) just like the tools
+      // runtime does — otherwise the hosted project boots empty and scorers that
+      // read remote state (e.g. the migration history) have nothing to assert on.
       await using hostedBackend = ev.metadata.hostedProject
         ? disposable(
             await bootPlatformBackend({
+              ...readSessionSeedArgs(ev),
               ref: HOSTED_PROJECT_REF,
               accessToken: HOSTED_ACCESS_TOKEN,
               hostname: "0.0.0.0",
--- a/apps/framework/harness/run-eval.ts
+++ b/apps/framework/harness/run-eval.ts
@@ -387,6 +387,9 @@ async function runOne(
               ref: HOSTED_PROJECT_REF,
               accessToken: HOSTED_ACCESS_TOKEN,
               hostname: "0.0.0.0",
+              // Expose Postgres-wire too, so linked DB workflows (`db push`,
+              // `migration repair`) reach the same project over the wire.
+              pgWire: true,
             }),
           )
         : undefined;
@@ -398,6 +401,7 @@ async function runOne(
           hosted: hostedBackend
             ? {
                 port: Number(new URL(hostedBackend.url).port),
+                pgPort: hostedBackend.pgPort,
                 ref: hostedBackend.ref,
                 accessToken: hostedBackend.accessToken,
                 mgmt: hostedBackend.mgmt,
--- a/apps/framework/harness/run-eval.ts
+++ b/apps/framework/harness/run-eval.ts
@@ -405,6 +405,7 @@ async function runOne(
                 ref: hostedBackend.ref,
                 accessToken: hostedBackend.accessToken,
                 mgmt: hostedBackend.mgmt,
+                query: hostedBackend.query,
                 invokeFunction: hostedBackend.invokeFunction,
               }
             : undefined,
--- a/evals/resolve-database-001-migration-history-mismatch/EVAL.ts
+++ b/evals/resolve-database-001-migration-history-mismatch/EVAL.ts
@@ -11,21 +11,10 @@ const ORPHAN_VERSION = "20240115000000";
 const PENDING_VERSION = "20240220000000";
 const SEEDED_PROFILE_COUNT = 25;
 
-/**
- * The exact local migration sequence a correct reconciliation must produce, in
- * order: the original `create_profiles`, the recovered orphan `add_profile_bio`
- * in its original slot, then the pending `add_avatar_url`. We pin the precise
- * set and order — not merely that timestamps ascend — because the orphan must be
- * recovered into its correct position (20240115, between create and avatar) so
- * local history fully describes the remote schema with no drift; discarding it
- * (repair --status reverted) or pulling it in under a fresh timestamp does not
- * qualify.
- */
-const EXPECTED_MIGRATIONS = [
-  { version: "20240101000000", name: "create_profiles" },
-  { version: ORPHAN_VERSION, name: "add_profile_bio" },
-  { version: PENDING_VERSION, name: "add_avatar_url" },
-] as const;
+// The two seeded migrations that bookend a correct reconciliation. The agent
+// must keep them, unrenamed, as the first and last migration.
+const FIRST_MIGRATION = { version: "20240101000000", name: "create_profiles" } as const;
+const LAST_MIGRATION = { version: PENDING_VERSION, name: "add_avatar_url" } as const;
 
 /**
  * Verifies recovery from a remote migration-history mismatch (AI-823): the
@@ -35,9 +24,14 @@ const EXPECTED_MIGRATIONS = [
  * Ground truth is read from the hosted project's database directly via
  * `ctx.hostedQuery` (in-process against the same PGlite, never the CLI under
  * test) plus the local migration files in the agent's workspace. We assert the
- * END STATE rather than the exact commands used, but the end state is pinned:
- * the local history must be exactly `EXPECTED_MIGRATIONS` (the orphan recovered
- * into its slot), and every one of those versions applied on the remote.
+ * END STATE rather than the exact commands used. Two recovered shapes are valid
+ * — both keep all three migrations, ordered create_profiles → bio reconciliation
+ * → add_avatar_url, with strictly ascending timestamps:
+ *   A. orphan recovered into its slot: bio at 20240115.
+ *   B. `db pull` style: bio captured under any timestamp, as long as it lands
+ *      before the avatar.
+ * `checkMigrationOrder` accepts either: create_profiles first, add_avatar_url
+ * last, exactly one migration strictly between, all applied on the remote.
  */
 const scorer: LocalStackScorer = async (ctx) => {
   try {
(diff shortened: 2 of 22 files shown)


evals/resolve-database-001-migration-history-mismatch/PROMPT.md
```diff
@@ -0,0 +1,19 @@
+---
+stage: resolve
+suite: benchmark
+interface: cli
+hostedProject: true
+projectRunning: false
```
mattrossman: Do we need project to be unstarted for this eval? In my test run, the agent stumbled at the start trying to invoke CLI commands before starting the stack, which doesn't seem essential for this scenario.

mattrossman: This exposed a possible bug, which is that `services:` is effectively ignored when `projectRunning: false` because the agent will just `supabase start` without filtering services.

This caused me to get warnings on stderror about mismatched service versions for versions we don't need/pin in this scenario:
```
WARNING: You are running different service versions locally than your linked project:\nsupabase/postgres:15.8.1.040 => supabase/postgres:17.6.1.064\nsupabase/gotrue:v2.177.0 => supabase/gotrue:v2.184.0\npostgrest/postgrest:v12.2.12 => postgrest/postgrest:v14.1\nsupabase/storage-api:v1.33.0 => supabase/storage-api:v1.33.0\nRun supabase link to update them.
```

apps/framework/harness/run-eval.ts
mattrossman: This `bootPlatformBackend` call isn't populating seed data (`projectSeedSql`, `logsSeedJsonl`, `functionsSeedDir`) so the migrations schema in `remote/project.sql` isn't taking effect in the run.

This caused the eval to fail in my test run:

```
  "checks": [
    {
      "name": "scorer evaluated migration-history recovery",
      "passed": false,
      "notes": "relation \"supabase_migrations.schema_migrations\" does not exist"
    }
  ],
```

evals/resolve-database-001-migration-history-mismatch/EVAL.ts
```diff
@@ -0,0 +1,245 @@
...
+async function checkHistoryReconciled(
+  ctx: LocalStackEvalContext,
+): Promise<CheckResult> {
+  const name = "remote migration history matches local migration files";
+  const remoteVersions = await remoteHistoryVersions(ctx);
+  const localVersions = await localMigrationVersions(ctx);
+  const orphans = remoteVersions.filter((v) => !localVersions.includes(v));
+  const orphanResolved = !remoteVersions.includes(ORPHAN_VERSION) ||
+    localVersions.includes(ORPHAN_VERSION);
+  const passed = orphans.length === 0 && orphanResolved;
```
mattrossman: nit, `orphans.length === 0` already verifies the orphan is resolved so that var is dead code

apps/framework/harness/run-eval.ts
Rodriguespn: Good catch, thanks. the hosted backend was booted without `projectSeedSql`, so `remote/project.sql` never loaded. Fixed

evals/resolve-database-001-migration-history-mismatch/EVAL.ts
Rodriguespn: Removed it along with `ORPHAN_VERSION` as well

evals/resolve-database-001-migration-history-mismatch/PROMPT.md
Rodriguespn: > Do we need project to be unstarted for this eval? In my test run, the agent stumbled at the start trying to invoke CLI commands before starting the stack, which doesn't seem essential for this scenario.

After trying a couple different ways, was able to reproduce that but I'm keeping `projectRunning: false`

The root cause for the old implementation was that the linked `supabase db push` couldn't actually reach the mocked hosted DB, and this confused the agent. In some runs it ran `supabase start` and reconstructed the scenario on that local stack as a decoy instead of the hosted remote (the scorer reads the remote).

The best solution I could think of was a CLI wrapper (the same seam that injects `-x` into `supabase start`). It now appends `--db-url <pooler-url>` to linked DB commands. (The CLI hardcodes the linked port to 5432)

Rodriguespn: > This exposed a possible bug, which is that `services:` is effectively ignored when `projectRunning: false` because the agent will just `supabase start` without filtering services.
> 
> This caused me to get warnings on stderror about mismatched service versions for versions we don't need/pin in this scenario:
> 
> ```
> WARNING: You are running different service versions locally than your linked project:\nsupabase/postgres:15.8.1.040 => supabase/postgres:17.6.1.064\nsupabase/gotrue:v2.177.0 => supabase/gotrue:v2.184.0\npostgrest/postgrest:v12.2.12 => postgrest/postgrest:v14.1\nsupabase/storage-api:v1.33.0 => supabase/storage-api:v1.33.0\nRun supabase link to update them.
> ```


Good catch, though it turned out to be the version cache, not service filtering. `services:` isn't actually ignored: the same CLI wrapper appends `-x <excluded>` to any agent-run `supabase start`, so the exclusions still apply.

The warnings came from the `.temp/*-version` files we seed (the linked project's versions) not matching the bundled CLI's stack. That comparison fires for every service regardless of `-x`. I've set those to match the bundled CLI versions, so no more mismatch. And in practice the agent no longer runs `supabase start` at all (it pushes straight to the remote), so it doesn't come up anyway

Rodriguespn: Follow-up on this, the scenario now has `projectRunning: true`, so the agent starts with the stack already up.

On the ports: the CLI hardcodes the linked/remote DB to 5432, but platform-lite runs on an ephemeral port for per-eval isolation, so a plain linked `db push` can't reach it.

So I added a thin CLI shim that shadows `supabase` on PATH. It only appends `--db-url <pooler-url>` to linked DB commands (`db push`/`db pull`/`db dump`/`migration repair`/`migration list`) and `-x <excluded>` to `supabase start`; everything else execs the real CLI untouched.

evals/resolve-database-001-migration-history-mismatch/PROMPT.md
```diff
@@ -0,0 +1,19 @@
...
+projectRunning: true
+services: []
+product:
+  - database
+topic:
+  - migrations
+motivation: AI-823
+---
+
+I added a migration that gives profiles an avatar and I'm trying to ship it to our hosted project, but `supabase db push` keeps failing with "The remote database's migration history does not match local files."
```
mattrossman: I feel like we could be less forgiving here to boost signal of results, some of these things I don't think users should have to explain.

For example "without resetting or wiping the hosted database, it holds real user profiles" - I expect agents to exercise caution on hosted data w/o needing to be told, I'd just keep a check for this to make sure agents aren't nuking it in the resolution process.

Or the context on how the migration conflict occurred and how they expect it to be resolved, my hope is that agents can figure what the conflict is on their own and suggest or take reasonable resolution strategies because users might not know or understand this themselves.

mattrossman (approved):
Tested this working locally, scenario looks good overall. I can't claim to understand all the pg-wire details, but it looks reasonable given the inherent complexity.

One non-blocking comment about making the scenario a bit more ambiguous to give us extra signal about agent weaknesses: https://github.com/supabase/evals/pull/43/changes#r3509413784

evals/resolve-database-001-migration-history-mismatch/PROMPT.md
mattrossman: Even the error message could be trimmed, as a user I'd probably be lazy and say something like "I'm trying to a ship a migration and it's not working" and let the agent discover the error + cause on its own.

Rodriguespn: Thanks for the suggestion, I took it all the way. The prompt is now a vague, realistic user report ("trying to ship a migration and it's not working"): no error text, no explanation of the cause, and no "don't wipe the data" hint.

One thing surfaced while testing with `gpt-5.4-mini`: with the vaguer prompt it ran `supabase link` as a sanity check and hit a DNS error (`host.docker.internal` doesn't resolve under the CLI's default resolver). 
Rather than mock out the full `link` command, I wrapped the CLI binary to force `--dns-resolver native` on `link`, same shim pattern we already use to inject `--db-url` on `db push`. After that the run went from 1/6 to 5/6, with the only failure being the method guard correctly catching the agent applying the migration via direct SQL instead of the CLI, exactly the kind of weakness the harder prompt is meant to expose

PR #45, opened 2026-06-18
feat(sandbox): install agent skills into the local-stack sandbox (AI-845)

Rodriguespn:
Install the experiment's agent skills into the local-stack sandbox with Vercel's `skills` package and expose them via progressive discovery (name+description in the prompt; the agent reads `.claude/skills/<name>/SKILL.md` on demand), while tools mode keeps injecting full skill text into the prompt.

Closes AI-845

### DIFF
--- a/apps/framework/harness/run-eval.ts
+++ b/apps/framework/harness/run-eval.ts
@@ -12,7 +12,12 @@ import {
 } from "node:fs";
 import { join, dirname } from "node:path";
 import { fileURLToPath, pathToFileURL } from "node:url";
+import { jsonSchema, tool, type ToolSet } from "ai";
 import { parseEvalMarkdown } from "@supabase-evals/core/eval-markdown";
+import {
+  frontmatterDescription,
+  stripFrontmatter,
+} from "@supabase-evals/sandbox";
 import {
   normalizeExperimentName,
   readRepeatedFlag,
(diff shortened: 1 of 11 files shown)


apps/framework/harness/run-eval.ts
```diff
@@ -135,6 +136,11 @@ function discoverEvals(): EvalManifest[] {
   return out;
 }
 
+/**
+ * Tools-mode skill loading: the agent has no filesystem, so each skill's full
+ * SKILL.md is injected into the system prompt. (Local-stack mode installs the
```
mattrossman: To keep things consistent, maybe both modes should either (a) inject the full skill in system prompt or (b) only inject the name/description and let the agent load read skill lazily.

For tool mode, we could use this `loadSkillTool` pattern:
https://ai-sdk.dev/cookbook/guides/agent-skills#step-4-create-the-load-skill-tool

mattrossman (approved):
Tested by running an eval locally, saw agent reading skill files as expected

Image

Just one comment here about how we might make this lazy loading behavior more consistent in tool mode evals: https://github.com/supabase/evals/pull/45/changes#r3437616676

apps/framework/harness/run-eval.ts
Rodriguespn: I'm leaning to implement `loadSkillTool` for tool mode to match the progressive disclosure property of skills

Rodriguespn: Done, thx for the suggestion

PR #47, opened 2026-06-18
feat(eval): add Claude Code CLI agent harness

Rodriguespn:
Adds **`claude-code`** as an eval agent alongside `ai-sdk`, so scenarios run through a **real agent harness** — the Claude Code CLI drives its own loop, tools, and MCP client — rather than a bare model loop. Its transcript is parsed back into the same surface scorers already use, so scoring is unchanged. The framework is generic: a runner + a parser + a registry entry adds the next agent (Codex follows in [#49](https://github.com/supabase/evals/pull/49)). 

## How it runs
The agent runs in a Docker sandbox in **both** modes. The base sandbox carries the agent's own tooling (shell/file tools) and skills, but **not** the Supabase CLI. The only thing that varies by mode is the Supabase local stack:

- **local-stack mode** — the Supabase CLI is installed at setup and the local stack is started (`supabase start`); the agent works with the CLI + its own tools, scored against the running stack.
- **tools mode** — **no Supabase CLI and no local stack**; the agent works through the Supabase **MCP** against platform-lite (the mocked hosted platform), with its shell/file tools still available.

**Main difference: tools mode has no Supabase CLI (or local stack)** — it's MCP-driven by environment, not by a tool restriction.

## Live runs
- [claude-code-haiku-4.5](https://github.com/supabase/evals/actions/runs/27825575802/job/82349020404)
- [claude-code-sonnet-4.6](https://github.com/supabase/evals/actions/runs/27825575802/job/82349020420)

Closes [AI-848](https://linear.app/supabase/issue/AI-848)


apps/web/vercel.json
```diff
@@ -0,0 +1,4 @@
+{
```
Rodriguespn, line 1: Needed to manually force a redeploy of the preview page

Note: This only triggers a redployment if the top commit changes `evals-results.json` (like running `run-evals` workflow)

packages/core/src/parsers/claude-code.ts
```diff
@@ -0,0 +1,239 @@
...
+    if (content) {
+      events.push({ timestamp, type: "message", role: "system", content, raw: data });
+    }
+  } else if (type === "result") {
+    // Terminal line of `claude --print`: carries the final assistant text and a
+    // subtype (`success`, `error_max_turns`, …). Emit the text as the closing
+    // assistant message so report derivation works off events alone.
+    const result = typeof data.result === "string" ? data.result : undefined;
+    if (result) {
+      events.push({ timestamp, type: "message", role: "assistant", content: result, raw: data });
```
mattrossman, line 205: Does this double-add the final assistant message? Since we have a
```js
 if (content) {
    events.push({ timestamp, type: "message", role: "assistant", content, raw: data });
  }
```
standalone block earlier

apps/framework/harness/run-eval.ts
```diff
@@ -444,23 +451,38 @@ async function runOne(
...
-    const session = await exp.runtime.startSession(readSessionSeedArgs(ev));
+    // Tools mode: the eval's tool surface is MCP (platform-lite). A CLI agent
+    // gets the same sandbox as local-stack minus the running stack — with its
+    // skills installed — and reaches the in-container MCP servers' host-side
+    // platform-lite via host.docker.internal (so platform-lite binds 0.0.0.0).
+    // An in-process agent runs host-side with no sandbox.
+    const cliSandbox = agentRunsInSandbox
+      ? await createBareSandbox({ skills: skillSources })
+      : undefined;
+    const session = await exp.runtime.startSession({
```
mattrossman: Now that there are 2 potentially `throw`-ing async ops here, there's opportunity for resource leak (if creation of the `session` fails, the `cliSandbox` won't reach the `finally` cleanup step).

Maybe we should implement `Symbol.asyncDispose` for these resources so we can [`await using`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/await_using) them so their lifetime is more predictable

apps/framework/harness/run-eval.ts
mattrossman: Not new from this PR, but noting that [similar to this comment](https://github.com/supabase/evals/pull/47/changes#r3460330496) there's a potential cleanup issue we might want to address w/ `await using` pattern

mattrossman (approved):

apps/web/vercel.json
Rodriguespn: Fixed

PR #49, opened 2026-06-18
feat(eval): add Codex agent & add `reasoningEffort` as experiment param

Rodriguespn:
Adds **`codex`** (OpenAI Codex) as an eval agent alongside `claude-code` and `ai-sdk`, built on the runner/parser framework from [#47](https://github.com/supabase/evals/pull/47): a runner + a parser + a registry entry. The orchestration is unchanged and Codex's transcript is parsed back into the same surface scorers already use, so scoring is unchanged.

Codex runs headless via **`codex exec --json`** in a Docker sandbox. Like Claude Code, it runs in **both** modes — the sandbox carries its shell/file tools either way; tools mode just drops the Supabase CLI + local stack, so Supabase access goes through **MCP**. Which mode an eval uses is a property of the eval (`interface`/`local/` dir), not the agent.

- Install + `codex login --with-api-key` (key piped via stdin, never in argv).
- Prompt (system + user) on stdin; `--skip-git-repo-check --dangerously-bypass-approvals-and-sandbox -m <model>`.
- MCP servers written to a generated `~/.codex/config.toml`.
- `codex exec` exits `0` even on a failed turn, so the stop reason is derived from the terminal `turn.completed` / `turn.failed` event rather than the exit code.

## Important
Codex enables a `tool_search` tool that small models reject — `gpt-5.4-nano` returns `400 invalid_request_error: "Tool 'tool_search' is not supported with gpt-5.4-nano."` and fails the turn immediately. **Use `gpt-5.4-mini` or larger**; the committed experiment uses `gpt-5.4`.

Closes [AI-849](https://linear.app/supabase/issue/AI-849)


experiments/codex-gpt-5.4.ts
```diff
@@ -0,0 +1,22 @@
...
+} from "@supabase-evals/core";
+import { localStackRuntime } from "@supabase-evals/sandbox";
+
+// Codex runs in both modes, like Claude Code: `runtime` drives tools-mode evals
+// (the runner writes its MCP servers into ~/.codex/config.toml against platform-
+// lite) and `localStack` drives local-stack evals. Which mode an eval uses is a
+// property of the eval (interface/local dir), not the agent.
+export default defineExperiment({
+  agent: codexAgent({
+    model: "gpt-5.4",
```
mattrossman: Per sync, I'd remove this experiment or replace w/ 5.4-mini to keep experiments reflective of minimal clusters of users. I expect power users will use 5.5, and those on a token budget will use 5.4-mini

mattrossman: Also, I don't see a way to specify reasoning level on `codexAgent` (or the claude code one).

Maybe that can be a separate PR but I think we need a way to control reasoning level in Codex and Claude Code agents since that's a pretty common setting users may change in those agents.

mattrossman (approved):
Looks good + tested locally, though I'd like to remove or replace the 5.4 experiment. If replacing, I'd pick 5.4-mini (medium) to mirror sonnet tier. Eventually will want a way to control reasoning explicitly though that can be a follow up PR.

experiments/codex-gpt-5.4.ts
mattrossman: see motivation on model selection here https://supabase.slack.com/archives/C0BAGJBL49E/p1782333782523379

PR #50, opened 2026-06-19
fix(evals): correct false-negative scoring (seed dir, endpoint case, brittle test check)

Rodriguespn:
Fixes false-negative eval scoring found while auditing [claude-code-haiku-4.5](https://github.com/supabase/evals/actions/runs/27825575802/job/82349020404) and [claude-code-sonnet-4.6](https://github.com/supabase/evals/actions/runs/27825575802/job/82349020420) runs. 

### Fixed
- seed never applied in `investigate-auth-001-deleted-user-access` and `investigate-realtime-001-subscribed-no-events` because their seed was still under the old `seed/` driectory (renamed to `remote/`. So both ran against an empty DB (scorer threw `relation "profiles" does not exist`; the realtime judge penalized the agent for rebuilding the scenario it was forced to recreate).
- `build-cli-002`'s "supabase db diff used" check matched `tc.endpoint === "bash"`, but Claude Code emit `"Bash"` instead. Lowercased `tc.endpoint`
- `discoverStackConfig()` read the API keys once and threw if absent. Now retries briefly (5×2s) before failing.
- The agent wrote passing tests and correctly diagnoses `posts` still failed it, because the "negative tests catch the bug (`failedTests > 0`)" check hard-coded one test style contradicted the diagnosis judge. Dropped it and kept only the judge assessment.
- the mock edge runtime's `require` shim only resolved `@supabase/supabase-js` and threw "import not supported" for `npm:`/`node:`/`esm.sh` imports the real Deno runtime accepts (`build-functions-004` used `npm:jwt-decode`). Now resolves `node:` builtins, `npm:`/`jsr:`/CDN URLs/bare specifiers (version-stripped) against the runtime's modules; unresolvable packages error clearly instead of being rejected. Verified via `bootPlatformBackend` (`npm:zod` + `node:crypto` → 200).

**Notes**
These are framework/eval-definition fixes, independent of the Claude Code feature work (#47).


mattrossman (approved):
Down the road we may want to rethink the edge function JS imports stuff, support for `@supabase/supabase-js` was a short term hack and I wonder how far we can get w/ this solution of installing everything the agent might need in our evals repo deps. Maybe there's a middle ground between that and spinning up the full edge functions runtime in local stack mode, like running actual Deno + letting it resolve deps with any version.

Rodriguespn:
> Down the road we may want to rethink the edge function JS imports stuff, support for `@supabase/supabase-js` was a short term hack and I wonder how far we can get w/ this solution of installing everything the agent might need in our evals repo deps. Maybe there's a middle ground between that and spinning up the full edge functions runtime in local stack mode, like running actual Deno + letting it resolve deps with any version.

Answered [here](https://supabase.slack.com/archives/C0BAGJBL49E/p1782143699725299?thread_ts=1781872770.064489&cid=C0BAGJBL49E)

PR #35, opened 2026-06-16
feat(benchmark): add pg_cron + pgmq queue workflow eval

mattrossman:
Adds a benchmark `build-database-001-pg-cron-queue-workflow` eval for wiring up a `pg_cron + pgmq` pipeline: the agent must schedule a cron job that enqueues tasks into a pgmq queue, then deploy an Edge Function to drain it.

**What the scenario is**

Seeds a Supabase project with `pg_cron` and `pgmq` already enabled. The agent is asked to schedule an `enqueue-tasks` cron job to run every minute and push tasks into a `tasks` queue, then deploy a `process-tasks` Edge Function that an internal worker can call.

**What the eval checks**
- `enqueue-tasks` cron job exists in `cron.job`
- The cron command actually executes and enqueues to the `tasks` queue
- The `process-tasks` function dequeues and returns messages

**Supporting fixes**

- Stub `pg_cron` and `pgmq`/`pgmq_public` schemas added to platform-lite. PGlite can't load the real compiled extensions, so these are implemented in pure SQL to cover the API surface needed for scoring
- `CREATE EXTENSION pg_cron/pgmq` intercepted as a no-op in the database routes (stubs are always-on)
- `cron` and `queues` added to the eval product enum
- Eval discovery now skips directories missing a `PROMPT.md` to avoid crashing on stale directories
- `@supabase/lite` upgraded to `0.4.1-next.4`; `pgmq_public` wrappers now support schema-qualified Supabase RPC calls in platform-lite

**Results**

The common failure mode is agents writing the edge fn with `deno.land/x/postgresjs` instead of using the Supabase SDK as shown in our guides. While it may work in the wild, it (a) isn't supported in platform-lite and (b) isn't following our guides, so I'm considering it a check failure. We may need to tune our docs search to better surface the relevant docs pages. In one case the agent used `@supabase/server` which is also unsupported in our lite edge fn runtime. If this continues to be a pain point we can turn this into a CLI scenario for better edge fn compatibility.

**References**
- https://supabase.com/docs/guides/queues/consuming-messages-with-edge-functions
- https://supabase.com/docs/guides/queues/quickstart#enqueueing-and-dequeueing-messages
- https://github.com/pgmq/pgmq#sql-only
- https://www.npmjs.com/package/@supabase/lite/v/0.4.1-next.4

Closes AI-812

PR #52, opened 2026-06-22
feat(benchmark): pg_cron + pgmq queue workflow as a local CLI eval

mattrossman:
Adds `build-cli-003-pg-cron-queue-workflow`, a CLI / local-stack alternative to #35.

Runs against the local stack so `pg_cron`/`pgmq` and the Edge Function runtime are all real instead of mocked in platform-lite. Seeds the `pgmq_public` wrappers from [Studio's source](https://github.com/supabase/supabase/blob/0433eeb5f5cc6c41b80c85793a3b7be1fc976119/packages/pg-meta/src/sql/studio/database/queues.ts) since there's no CLI command to create them locally (on hosted it's the "Expose Queues via PostgREST" toggle).

`process-tasks` drains via `pgmq_public` the way the consuming-with-edge-functions guide shows (a scheduled worker that reads messages and clears them). Note gating the function to the service role isn't tested here, partly b/c the new `@supabase/server` auth modes (`auth:'secret'`/`'user'`) don't work on the local stack yet (the runtime only injects the legacy keys, no secret-key env or JWKS).

**Checks**
- `enqueue-tasks` runs every minute
- its command enqueues to `tasks`
- `process-tasks` drains the queue

gpt-5.5 passes 3/3, draining via `pgmq_public`.

**References**
- https://supabase.com/docs/guides/cron/install
- https://supabase.com/docs/guides/queues/pgmq
- https://supabase.com/docs/guides/queues/consuming-messages-with-edge-functions
- https://github.com/supabase/supabase/blob/0433eeb5f5cc6c41b80c85793a3b7be1fc976119/packages/pg-meta/src/sql/studio/database/queues.ts (the `pgmq_public` wrappers we seed)

Closes AI-812


evals/build-cli-003-pg-cron-queue-workflow/EVAL.ts
```diff
@@ -0,0 +1,176 @@
...
+      passed: false,
+      checks: [{ name: "scorer completed without errors", passed: false, notes: msg }],
+    };
+  }
+};
+
+export default scorer;
+
+/** Run arbitrary SQL against the local stack as postgres. base64 sidesteps shell quoting. */
+async function execSql(ctx: LocalStackEvalContext, sql: string): Promise<CommandResult> {
```
mattrossman, line 41: Currently `ctx.query()` only supports reads hence this raw SQL execution helper, but eventually I'd like the `ctx.` helper to be more general purpose like our `execute_sql` MCP tool. Not adding to this PR to avoid conficts w/ other ongoing work.

PR #54, opened 2026-06-23
feat(ci): add run-evals-changed label to run only changed evals

mattrossman:
Adds a `run-evals-changed` label that runs only the evals whose directories were modified in the PR, instead of the full suite. This prevents wastefully re-running every "eval x experiment" permutation when drafting a PR.

Leverages the existing `eval-refresh.yml` workflow to support optional matrix filtering by touched `evals/<id>/` dirs via `gh pr diff`. If no eval dirs changed, the workflow skips gracefully.

Adds a `--merge` flag to `export-results` that splices new results into the existing JSON rather than overwriting it to support partial refreshes.

**Demo**

I tested adding the new and existing labels to this PR to verify new label works and existing didn't regress.

`run-evals-changed`: only `resolve-dataapi-001-empty-results` ran ([run](https://github.com/supabase/evals/actions/runs/28000281773)). The results commit updated only that eval's entry in `eval-results.json` ([commit](https://github.com/supabase/evals/pull/54/commits/10cdfd9b533f617e6e3615d5eb4da27f8606712c)).

`run-evals`: full suite ran (28 matrix jobs), all passed, results commit touched evals across the file as expected ([run](https://github.com/supabase/evals/actions/runs/28000638348), [commit](https://github.com/supabase/evals/pull/54/commits/c7ad77b914740ce650c687b1ea00903190b3be93)).

Closes AI-855

### DIFF
--- a/.github/workflows/eval-refresh.yml
+++ b/.github/workflows/eval-refresh.yml
@@ -43,9 +43,12 @@ jobs:
     if: >-
       github.event_name == 'workflow_dispatch' ||
       (github.event_name == 'pull_request' &&
-        contains(github.event.pull_request.labels.*.name, 'run-evals') &&
+        (contains(github.event.pull_request.labels.*.name, 'run-evals') ||
+         contains(github.event.pull_request.labels.*.name, 'run-evals-changed')) &&
         github.event.pull_request.head.repo.full_name == github.repository &&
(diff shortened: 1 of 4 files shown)


mattrossman:
Merged after syncing with Pedro

PR #70, opened 2026-06-26
feat: explicit suite on every eval

mattrossman:
Makes `suite` required on every eval (`benchmark` | `regression` | `other`) so future evals are accounted for, and tags the existing 31: 17 benchmark, 12 other (Saxon's bootstrap evals), 2 regression.

Experiment `suite` becomes a list so one config can run in multiple suites. `claude-code-sonnet-4.6` is now `["benchmark", "regression"]`, so `--suite regression` runs it against the regression eval set. A single `--suite X` selects experiments tagged X against evals declaring X.

Closes AI-859

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ pnpm web
 
 ### Add an experiment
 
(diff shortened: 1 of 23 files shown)


mattrossman:
Cancelling workflow as the only thing I wanted to check was that filtering still works as intended.

PR #72, opened 2026-06-29
feat: track skill activation in eval results

mattrossman:
**Changes**
- Adds `skills.available` and `skills.loaded` to eval result metadata, including result export, so new runs record which configured skills were exposed and which ones were actually loaded.
- Tracks loaded skills from the tools-mode `load_skill` tool, Claude Code `Skill` calls, and CLI tool calls that read `skills/<name>/SKILL.md`. Adds focused test coverage for the activation summary builder.

In testing, agents are pretty reliable loading the `supabase` skill, they don't often load `postgres-best-practices`.

Closes AI-874

### DIFF
--- a/packages/core/src/skill-results.ts
+++ b/packages/core/src/skill-results.ts
@@ -1,6 +1,7 @@
 import type { ToolCallRecord } from "./index.js";
 import type { SkillResult } from "./eval-metadata.js";
 
+// Skill reads can appear as bare paths or quoted shell args in tool-call logs.
 const PATH_START_BOUNDARY = String.raw`(?:^|[\s"'` + "`" + String.raw`])`;
 const PATH_END_BOUNDARY = String.raw`(?:$|[\s"'` + "`" + String.raw`])`;
 
--- a/packages/core/src/skill-results.ts
+++ b/packages/core/src/skill-results.ts
@@ -1,19 +1,16 @@
 import type { SkillResult } from "./eval-metadata.js";
 import type { ToolCallRecord } from "./index.js";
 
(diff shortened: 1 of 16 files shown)


packages/core/src/skill-results.ts
```diff
@@ -0,0 +1,50 @@
...
+export function buildSkillResult(
+  available: readonly string[],
+  toolCalls: readonly ToolCallRecord[],
+): SkillResult {
+  const loaded = available.filter((skill) =>
+    toolCalls.some((call) => callLoadsSkill(call, skill)),
+  );
+
+  return {
+    available: [...available],
```
gregnr: Double checking this shallow clone is intentional?

gregnr (approved):

Rodriguespn:
Left some comments on slack
https://supabase.slack.com/archives/C0BAGJBL49E/p1782835002807109?thread_ts=1782758051.905169&cid=C0BAGJBL49E

mattrossman:
Thanks for the suggestions @Rodriguespn, I've added an agent-agnostic `.loadedSkill` to tool call records so each agent owns the logic of how to populate it. The run-level `skills.loaded` result field is now derived from that normalized value.

packages/core/src/skill-results.ts
mattrossman: Just a quirk of these parameters being `readonly string[]` though there's not much reason for them to be `readonly`, simplified in https://github.com/supabase/evals/pull/72/changes/b7c2f72f78c375c0e42ff8c1bbb96e7e21ec4900 by making these regular mutable params and removing the clone.

PR #73, opened 2026-07-02
feat: bump Sonnet experiments to Claude Sonnet 5

mattrossman:
Swaps the three Sonnet 4.6 experiments to [Sonnet 5](https://www.anthropic.com/news/claude-sonnet-5) and updates the matching README references.

According to [Claude models overview](https://platform.claude.com/docs/en/about-claude/models/overview) this is a stable ID:

> Every Claude model ID is a pinned snapshot. Models with a date in the ID (for example, 20250929) are fixed to that specific release. Starting with the Claude 4.6 generation, model IDs use a dateless format that is also a pinned snapshot, not an evergreen pointer

Sanity-checked `claude-code-sonnet-5` against `resolve-dataapi-001-empty-results`: 7/7 checks passed.

Closes AI-881

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Agent-backed runs require the relevant provider key in `.env` (e.g. `OPENAI_API_
 Run one eval:
 
 ```bash
-pnpm eval -- --eval resolve-dataapi-001-empty-results --experiment claude-code-sonnet-4.6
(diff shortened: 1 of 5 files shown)


mattrossman:
Results look good, merging as this is a small change

PR #80, opened 2026-07-09
fix: declarative schema guidance and scoring

mattrossman:
Improvements for `build-cli-002-declarative-schema`

- Updates the agent-skills submodule to include declarative schema workflow guidance from https://github.com/supabase/agent-skills/pull/120.
- Fixes eval refresh single-eval handling (which caused [this failure](https://github.com/supabase/evals/actions/runs/29025007584)), and supports comma-separated eval IDs.
- Fixes `build-cli-002-declarative-schema` scoring to use normalized `.command` when testing if `supabase db diff` was called, and normalizes AI SDK agent's tool calls onto `ToolCallRecord.command`

See refresh run: https://github.com/supabase/evals/actions/runs/29025860898

Results now look more reasonable, where Codex stumbles without explicit declarative schema guidance (it tried editing the existing migration in-place + resetting db) but passes with the skill. Interestingly Claude gets it right even without skills.

| Without skills | With skills |
|--------|--------|
| CleanShot 2026-07-09 at 10 47 06@2x | CleanShot 2026-07-09 at 10 47 04@2x | 

Before merging this, we should merge + release https://github.com/supabase/agent-skills/pull/120

Ref AI-875

### DIFF
--- a/.github/workflows/eval-refresh.yml
+++ b/.github/workflows/eval-refresh.yml
@@ -8,7 +8,7 @@ on:
         required: false
         default: ""
       eval:
-        description: "Optional single eval id to run"
+        description: "Optional comma-separated eval ids to run"
         required: false
         default: ""
       suite:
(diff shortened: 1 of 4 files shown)


Rodriguespn (approved):
LGTM. Thx!

PR #82, opened 2026-07-10
feat: dual-auth edge function evals (RLS user vs service/secret) + @supabase/server coverage

mandarini:
Adds two local-stack evals covering dual-auth Supabase Edge Functions — one endpoint serving both an RLS-scoped **user** path and an RLS-bypassing **service (secret key)** path — split by role per the contributing guide: a **benchmark** eval that gates on agents *discovering* `@supabase/server` unprompted, and a **regression** eval that guards agents *using it correctly* when pointed at it.

## What kind of change does this PR introduce?

New evals (feature). No changes to existing evals' behavior.

### `build-functions-005-dual-auth-user-secret` — benchmark: adoption is gating

Requirements-only ticket (does **not** name `@supabase/server`, no inline schema — the `user_stats` table is discoverable in `supabase/migrations/`) to build and serve a `user-stats` Edge Function. `EVAL.ts` is a `LocalStackScorer` that signs up two users, seeds a row each, and invokes the deployed function over HTTP with different credentials (the new `sb_secret_…` / `sb_publishable_…` keys; per-eval `cliVersion: 2.109.1` so the edge runtime exposes the new-key env the package needs).

Gating checks:
- rejects a request with no credentials
- a user JWT returns only that user's own rows (RLS enforced)
- a user cannot escalate by passing another `user_id` in the body
- the secret key bypasses RLS to read the target user's rows
- a non-service key is not granted service access
- **a forged, unsigned (`alg:none`) token is rejected** (verifies the function actually validates JWT signatures, not just decodes them)
- **a user token placed in the `apikey` slot is not mistaken for the secret key**
- **the implementation uses `@supabase/server`** — a correct-but-hand-rolled solution fails

That last check is the point: earlier runs showed **0/8 agents reach for `@supabase/server` unprompted** (all hand-roll raw `supabase-js`) while passing every behavioral check. Gating on adoption gives the benchmark the failing-agent signal the contributing guide asks for, and points the fix at **docs/skills discoverability** — exactly the "reasonable attempt, fails for lack of context, correctable via docs/skills" shape we want to showcase. Expect this scenario to run red until discoverability improves; that is intentional.

### `build-functions-006-dual-auth-with-server` — regression: correct usage when named

Same scenario and checks, but the prompt **mandates** `@supabase/server` (`import { withSupabase } from "npm:@supabase/server"`). It sits in `suite: regression` as the usability canary: agents currently use the package correctly when pointed at it (**8/8**, including codex models), and this eval catches package or docs changes that break that. Runs nightly and in PR sweeps under the `regression` experiment suite.

## Findings recorded along the way

- **Discoverability, not usability, is the friction**: 0/8 spontaneous adoption vs 8/8 correct usage when named.
- A with/without-`supabase-server`-skill ablation (opus-4.8, sonnet-5) showed **no difference** — both models use the package correctly from the prompt naming it alone. The temporary `server` experiment suite used for that ablation was removed again in this PR.

## Additional context

- **Motivation:** `FDBKIN-19273` (service-role → secret-key / edge-function JWT confusion). The scorer drives the new API keys (`@supabase/server` rejects the legacy `service_role`/`anon` JWTs), and the migration grants table privileges explicitly since newer CLIs no longer auto-grant on migration-created tables.
- Runtime: local-stack (Docker + the real Supabase CLI); `projectRunning: false` — starting the stack is part of the task.


evals/build-functions-005-dual-auth-user-secret/PROMPT.md
```diff
@@ -0,0 +1,44 @@
...
+  - edge-runtime
+projectRunning: false
+motivation: FDBKIN-19273
+---
+
+Build and serve a Supabase Edge Function named `user-stats` for this project,
+reachable over HTTP at `/functions/v1/user-stats`.
+
+Our product stores per-user metrics in a `user_stats` table that already exists
+(see `supabase/migrations/`), protected by row-level security so a user can read
```
mattrossman: Do we need to specify this? I expect agents to read the project for context on what already exists

mattrossman (approved):

evals/build-functions-005-dual-auth-user-secret/PROMPT.md
```diff
@@ -0,0 +1,44 @@
...
+2. **Our internal billing service**, a trusted backend with no signed-in user.
+   It authenticates with the project's secret (service-role) key in the `apikey`
+   header, and names the target user with a `user_id` in the JSON request body.
+   It needs that user's stats.
+
+Return the matching rows as JSON. The endpoint must be secure: only ever serve
+stats to a caller that is genuinely entitled to them, and turn away callers that
+are not.
+
+Get the local stack running so the function is reachable at the path above.
```
mattrossman: Curious how important this is for the scenario vs having it already running, it seems like that would only make the eval slower to run / liable to fail for reasons unrelated to the main supabase/server usage thing we're testing

evals/build-functions-005-dual-auth-user-secret/PROMPT.md
mandarini: Agreed — trimmed to a bare mention of the table in 96ab2f1; the RLS/migrations description is discoverable from the project and the security requirement is already stated at the end.

evals/build-functions-005-dual-auth-user-secret/PROMPT.md
mandarini: Good call — our own data backs this: the one non-security failure so far (codex-gpt-5.4-mini, both attempts) was exactly stack startup, not the thing we're testing. Dropped the step in 96ab2f1 (projectRunning back to the default true) so the eval isolates the @supabase/server signal. Will do one more label run to validate before merging.

PR #87, opened 2026-07-14
fix(evals): accept secure 401/403 in service-role-bypass scorer

barryroodt:
## Summary

- Accept any graceful, non-leaking response to a forced cross-user read: `200` scoped to the caller, or a `401`/`403`/`404` denial. The exact status is left to the implementation; the no-leak invariant is the real contract.
- Keep the no-leak invariant for every response and retain the caller-JWT requirement for reads.
- Add response-path diagnostics (`status=...`, bearer-token count) without logging bearer token values.
- Dedupe the denial checks behind a single `secureResponse` helper so the accepted-status policy and note format are single-sourced (previously the two force-read checks were copy-paste twins).
- Add focused scorer smoke coverage for every accepted and rejected shape (AI-918 falsifiability).

## Original run evidence

The [committed result export at `12b223d`](https://github.com/supabase/evals/blob/12b223dd96315a418c40d8f9ab9e6306ff81fb72/apps/web/src/data/eval-results.json) for [run 28895076929](https://github.com/supabase/evals/actions/runs/28895076929) records the with-skills run at 3/5, with exactly these checks failing:

- `user A cannot force-read user B note`
- `user B cannot force-read user A note`

The no-skills result finished at 5/5. Its job log shows attempt 1 at 3/5 before retrying to 5/5, while both with-skills attempts remained at 3/5. This matches the issue's documented difference: the secure skills implementation returned `403`, while the passing no-skills implementation returned caller-scoped `200` responses.

Retries previously masked this scorer bug in the no-skills condition. With this fix, both secure behaviors are accepted deterministically, so retry masking no longer applies to this scenario.

## Falsifiability

The smoke regression proves both accepted behaviors (secure `401`/`403`/`404` denials and caller-scoped `200` reads) and still fails implementations that:

- return another user's note in either a `200` or a denial response,
- ignore missing authentication,
- return a `401` that leaks note data,
- use a service-role bearer token for the user-facing read.

## Verification

- `pnpm test:framework`: reaches `PASS service-role bypass scorer security contract` (all secure/insecure scenarios), 15.63s. The run then hits the pre-existing frontend smoke failure (missing `node_modules/vite/bin/vite.js`), which reproduces on clean `main`.
- `pnpm --filter @supabase-evals/framework typecheck`: PASS (1.94s)
- [Changed-eval run 29333039515](https://github.com/supabase/evals/actions/runs/29333039515) on the first commit: all matrix jobs PASS; refreshed export in [`d234af9`](https://github.com/supabase/evals/commit/d234af96f31b282acdfcb88cf14ec244dcc8ceb0) shows `codex-gpt-5.4-mini` and `codex-gpt-5.4-mini-no-skills` both PASS with no failed checks. A fresh run re-validates the 404/restructure commit.
- No local compile or test step exceeded one minute.

Closes AI-910

### DIFF
--- a/apps/framework/scripts/smoke-framework.ts
+++ b/apps/framework/scripts/smoke-framework.ts
@@ -4,7 +4,12 @@ import { dirname, join } from "node:path";
 import { fileURLToPath, pathToFileURL } from "node:url";
 import { bootPlatformBackend } from "../harness/platform-backend.js";
 import { viteBuild, vitestRun } from "../harness/project-runner.js";
-import type { ToolScorer, TranscriptPart } from "../harness/types.js";
+import type {
+  EdgeFunctionsInvokeResult,
+  ToolEvalContext,
+  ToolScorer,
+  TranscriptPart,
+} from "../harness/types.js";
 import type { PlatformBackend } from "../harness/platform-backend.js";
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ROOT = join(__dirname, "..", "..", "..");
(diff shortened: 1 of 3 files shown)


barryroodt:
Following up with one useful data point from the refreshed results:

**The no-skills run independently chose the same secure `403` behavior.** Both force-read checks report `status=403`, and the run passed 5/5 on its first attempt. Under the old scorer, that result would have failed despite preventing the leak.

The with-skills run also passed both `status=403` checks, finishing on its second attempt. This directly confirms the scorer change removes response-shape bias rather than favoring either condition.

mattrossman (approved):

PR #88, opened 2026-07-14
fix: unblock investigate-auth-001 — explicit re-login requirement, full auth helpers in platform-lite, fix-aware judge rubric

Rodriguespn:
Fixes the systematic failures in `investigate-auth-001-deleted-user-access` (1/8 benchmark runs passing) found while analyzing why loaded skills made results worse.

## Changes

- **`PROMPT.md`** — states outright that a deleted account must not be able to log back in. Agents were reasonably implementing soft-delete + session revocation while keeping the auth user, then failing `checkCannotSignBackIn`.
- **`packages/platform-lite` — create all four Supabase auth helper functions.** `@supabase/lite@0.5.1-next.1` never creates the auth helpers on our direct-exec init path, so eval projects only had the two we inlined ourselves (`auth.uid()`, `auth.role()`). Agent-written RLS policies using `auth.jwt()` failed with `function auth.jwt() does not exist`, diverging from production and pushing agents into fail-closed fallback policies. `AUTH_ROLES_SQL` now replicates all four helpers (`auth.uid()`, `auth.role()`, `auth.email()`, `auth.jwt()`), matching lite's `SUPABASE_AUTH_HELPERS_SQL` definitions.
- **`EVAL.ts` judge rubric** — the access-window criterion now judges the agent's claim against the fix it actually shipped instead of hardcoding one canonical answer: fixes that leave the Data API accepting stale JWTs must state the expiry window plus a mitigation; fixes that genuinely close the data path (e.g. RLS validating the user/session still exists) may say so, but every answer must still show that tokens stay locally valid until expiry (`getClaims()`-style validation).

> [!NOTE]
> The auth helpers are a temporary replication. Once the next supalite version ships with the auth helpers exported/applied on init ([LITE-287](https://linear.app/supabase/issue/LITE-287/auth-helpers-missing-from-direct-postgres-init-path), [Slack thread](https://supabase.slack.com/archives/C088MLLE0KU/p1783702603133869)), replace the custom helpers with the ones from supalite.

## Verification

- `pnpm typecheck` + `pnpm check` clean; platform-lite tests 38/39 (the one failure is the pre-existing `routes supabase-js rpc calls to PostgREST` supalite quirk, reproducible on a clean tree).
- `codex-gpt-5.4-mini` on `investigate-auth-001`: all 6 deterministic checks pass (previously the agent's session-validating RLS erred out); the judge fails only on the genuinely missing local-validation caveat — the discrimination the eval intends.

Close AI-909

### DIFF
--- a/evals/investigate-auth-001-deleted-user-access/PROMPT.md
+++ b/evals/investigate-auth-001-deleted-user-access/PROMPT.md
@@ -16,7 +16,7 @@ morning that same person was back: still signed in, reading and saving their
 data like nothing happened.
 
 Figure out why the account still works, fix the flow so a deleted account
-must not be able to log back in, and tell me whether there is any window where they could still
+loses access, and tell me whether there is any window where they could still
 get in after the fix.
 
 One more thing while you're at it: we're migrating off the legacy
--- a/evals/investigate-auth-001-deleted-user-access/PROMPT.md
+++ b/evals/investigate-auth-001-deleted-user-access/PROMPT.md
@@ -16,7 +16,7 @@ morning that same person was back: still signed in, reading and saving their
 data like nothing happened.
 
 Figure out why the account still works, fix the flow so a deleted account
-loses access, and tell me whether there is any window where they could still
+must not be able to log back in, and tell me whether there is any window where they could still
 get in after the fix.
 
 One more thing while you're at it: we're migrating off the legacy
--- a/apps/framework/harness/project-runner.ts
+++ b/apps/framework/harness/project-runner.ts
@@ -52,7 +52,7 @@ function setupSource() {
 import { existsSync, readFileSync, readdirSync } from "node:fs";
 import { join } from "node:path";
 import { afterAll } from "vitest";
-import { App, getAuthSchemaSql } from "@supabase/lite";
+import { App, getAuthSchemaSql, SUPABASE_AUTH_HELPERS_SQL } from "@supabase/lite";
 import { createPgliteConnection } from "@supabase/lite/pglite";
 
 const PROJECT_DB_URL = "http://supabase-evals.local";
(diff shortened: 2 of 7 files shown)


evals/investigate-auth-001-deleted-user-access/PROMPT.md
```diff
@@ -16,7 +15,7 @@ morning that same person was back: still signed in, reading and saving their
 data like nothing happened.
 
 Figure out why the account still works, fix the flow so a deleted account
-loses access, and tell me whether there is any window where they could still
+must not be able to log back in, and tell me whether there is any window where they could still
```
mattrossman: Tempted to leave this ambiguous, because IMO "a deleted account loses access" implies that the user shouldn't just be able to log back in. Like if there was a spam account abusing a service, revoking their session isn't really causing them to lose access to the service. Even if agents don't go so far as deleting the user I'd hope they at least flag it for human clarification.

Rodriguespn: Agree, thx for the suggestion. I made the prompt more ambiguous.

Note: I also updated the agent-skills submodule and upgraded the supalite package to `v0.7.1-next.3` ([Slack thread](https://supabase.slack.com/archives/C088MLLE0KU/p1784049781502659?thread_ts=1783702603.133869&cid=C088MLLE0KU)), which now exports `SUPABASE_AUTH_HELPERS_SQL` and `getStorageSchemaSql` so we no longer duplicate them. 
The migration looked low risk because a) the exported auth helpers are byte-identical to the copies we were inlining, b) the only notable behavioral changes don't affect this suite's evals

Happy to move the agent skills submodule update and the supalite package update to a separated PR

mattrossman (approved):

PR #91, opened 2026-07-15
chore: repo-wide Biome formatting (AI-831)

barryroodt:
Closes AI-831.

Adds repo-wide formatting tooling (Biome) with a top-level `format` script to cut formatting noise, mirroring the sibling `supabase/mcp` setup.

## Commits
1. **`chore: add biome formatting tooling`** — `biome.json` (formatter-only: linter/organizeImports off, single quotes, es5 trailing commas, space indent), root `format`/`format:check` scripts, `@biomejs/biome@1.9.4` dev dep, `apps/web` `format:check`, and a Code Quality CI workflow (`pnpm format:check` on PRs).
2. **`style: apply repo-wide biome + prettier formatting`** — mechanical `pnpm format` pass across 148 files. No behavioral changes; `pnpm typecheck` passes.

## Notes
- `apps/web` stays on Prettier: its `prettier-plugin-tailwindcss` class sorting has no Biome equivalent, so removing it would regress. The top-level `format` orchestrates Biome + the web app's Prettier, so it remains one repo-wide command.
- Biome ignores `dist`, `results`, `submodules`, `apps/web`, and the generated `management-api/types.ts`.
- CI uses the repo's pinned pnpm + node setup (`.node-version`) so it runs the pinned Biome 1.9.4, rather than mcp's `setup-biome` action.

## Verification
- `pnpm format:check` clean (Biome + web Prettier legs).
- `pnpm typecheck` passes.


biome.json
```diff
@@ -0,0 +1,41 @@
...
+      "submodules",
+      "packages/platform-lite/src/management-api/types.ts"
+    ]
+  },
+  "formatter": {
+    "enabled": true,
+    "indentStyle": "space"
+  },
+  "organizeImports": {
+    "enabled": false
```
gregnr, line 24: Any reason to make this false?

biome.json
```diff
@@ -0,0 +1,41 @@
...
+  },
+  "linter": {
+    "enabled": false
```
gregnr, line 27: Confirming this is intentional too?

barryroodt: Also intentional, for the same two reasons: it mirrors mcp's config, and linting felt like bigger scope than the formatting ticket. Enabling the recommended rules today surfaces 87 diagnostics, including non-mechanical findings such as 33 non-null assertions and 19 explicit `any`s that need fixes reviewed as real code changes, not a formatting pass. Worth noting `apps/web` keeps its own eslint and is ignored by Biome. I think turning the linter on deserves its own issue/PR, but lmk if you'd rather see it bundled here.

biome.json
barryroodt: Yeah, intentional: this config is copied verbatim from [mcp's biome.json](https://github.com/supabase/mcp/blob/10af00bbce7ff1dc116be36a78670119220fb2da/biome.json), which also keeps organizeImports off, and I wanted to keep this PR strictly formatting per AI-831. Enabling it flags 72 more files for import re-sorting on top of an already mechanical 100-file diff. It's fully auto-fixable (`biome check --write`) though, so if you'd rather have it on I can flip it here or in a small follow-up, either works for me.

biome.json
```diff
@@ -0,0 +1,41 @@
...
+    "useIgnoreFile": false
+  },
+  "files": {
+    "ignoreUnknown": false,
+    "ignore": [
+      "**/dist",
+      "apps/web",
+      "results",
+      "submodules",
+      "packages/platform-lite/src/management-api/types.ts"
```
mattrossman:
```suggestion
      "packages/platform-lite/src/management-api/types.ts",
      "packages/platform-lite/src/management-api/openapi.json"
```

Another generated file we may want to ignore

barryroodt: Good catch, added in 6eda0f4. Also reverted the format pass's churn on that file so the next `generate:types` run stays diff-free.

mattrossman (approved):
The biome/prettier split seems like a reasonable approach, I'm also happy to use prettier everywhere if that simplifies things but no strong preference.

PR #97, opened 2026-07-17
fix: `deploy-database-001-prometheus-metrics` file and MCP visibility

mattrossman:
Reduces result instability in `deploy-database-001-prometheus-metrics`'s scorer due to file-visibility gap.

The judge only read 3 hardcoded paths (`prometheus.yml`, `docker-compose.yml`, `README.md`), so a valid [`file_sd_configs`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config)-based scrape target split into its own yaml file (e.g. `supabase-targets.yml`) was invisible to the judge and got auto-failed as unverifiable. Scorer now reads every yaml file under `observability/` instead of a fixed set of hardcoded paths, with a test for the yaml collector to confirm it finds split target files and leaves `secrets/` and non-yaml files alone

Verified against the real GH Actions transcript from run https://github.com/supabase/evals/actions/runs/29531871411: the judge's "target wiring cannot be verified" note is gone, replaced with a substantive verdict on the actual placeholder value.

The eval was also missing `hostedProject: true`, so agents had no real hosted project to discover a project ref from and fell back to placeholders (`<project-ref>`, `__SUPABASE_PROJECT_REF__`), which the judge correctly failed as undeployable. That was the actual cause behind most of the placeholder-ref failures, not judge non-determinism. Added it and refreshed results again: 6/8 experiments now pass with real project refs resolved. The 2 remaining failures (`codex-gpt-5.4-mini`, `codex-gpt-5.4-mini-no-skills`) are genuine model mistakes, missing Secret API key creation steps in the README, and one using inline `basic_auth.password` instead of `password_file`. Not an eval bug.

Closes AI-911

### DIFF
--- a/apps/web/src/data/eval-results.json
+++ b/apps/web/src/data/eval-results.json
@@ -820,12 +820,12 @@
       {
         "name": "configured the Supabase Metrics API scrape correctly",
         "passed": false,
-        "judgeNotes": "Supabase scrape uses placeholder YOUR_PROJECT_REF rather than a real project ref, so it is not deployable as-is. Other required elements are present."
+        "judgeNotes": "Supabase scrape uses HTTPS, the correct metrics_path, Basic Auth with password_file, preserves the app job, and mounts the secrets directory. However, the target is still the placeholder `<PROJECT_REF>.supabase.co:443`, not a real project ref, so it is not deployable as-is."
       },
       {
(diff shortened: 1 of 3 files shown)


mattrossman:
Superceded by https://github.com/supabase/evals/pull/144

PR #99, opened 2026-07-21
platform-lite: serve the ClickHouse logs endpoint (/analytics/endpoints/logs)

barryroodt:
## What

Teaches platform-lite the ClickHouse logs endpoint that current mcp actually calls: `GET /v1/projects/{ref}/analytics/endpoints/logs`, taking ClickHouse-dialect SQL over the unified `logs` stream.

## Why

Since supabase/mcp#326, `get_logs` (and the proposed `query_logs` in supabase/mcp#333) query `/analytics/endpoints/logs` with ClickHouse SQL. platform-lite only served the legacy BigQuery-era `logs.all`, so **any logs eval against a locally built mcp 404s at the fixture**. The gap is masked today because evals pin a published `MCP_SERVER_VERSION`; it bites the moment anyone points the harness at an mcp checkout (which is how we validated mcp#333).

## How

- **Unified `logs` VIEW** over the existing seeded tables: a `source` discriminator plus a jsonb `log_attributes` map built from the flat columns, with seeded `metadata` as fallback. ClickHouse-shaped SQL runs against it with minimal translation.
- **Minimal, observed-only dialect translation** (`compileClickHouseLogsSql`): `log_attributes['k']` to jsonb access (numeric cast for status/exec-time keys so `>= 500` comparisons work) and `countIf(...)` to `count(*) FILTER (WHERE ...)`. A small shim family (`toInt32OrZero`/`toInt64OrZero`/`toUInt32OrZero`/`toString`) covers casts models genuinely emitted during live runs. Anything else surfaces the raw SQL error to the model, which is deliberate: the supported surface is documented and only grows from observed model output.
- **Read-only is enforced by a postgres read-only transaction**, not regex: mutating SQL, including data-modifying CTEs (`WITH x AS (DELETE ...) SELECT`), is rejected before it can touch shared fixture state.
- **`iso_timestamp_start`/`end` are accepted but ignored**, matching the legacy route: scenario seeds carry fixed dates while mcp defaults windows from the current clock, so a faithful filter would empty every scenario. Documented in-code as a known limitation; window-correctness needs relative-time seeding and a discriminating eval (follow-up).
- Contract tests use **verbatim SQL captured from live claude-sonnet-5 runs** (the mcp `edge-function` preset, a `countIf` aggregation, and the exact `toInt32OrZero(toString(...))` query the model emitted), plus the runtime source and the legacy route untouched.

## Verification

- `pnpm typecheck` clean, `pnpm vitest run src/management-api/debugging.test.ts` 8/8: five translator/view contract tests plus three route-level tests at the HTTP boundary (normal ClickHouse query returns the `{result}` shape; `WITH x AS (DELETE ... RETURNING *) SELECT` is rejected by the read-only transaction with fixture rows asserted unchanged; plain non-SELECT hits the 400 prefix gate).
- Live: `investigate-logs-001-top-error-function` passes against a locally built mcp `main` (3/3 checks) where it previously 404'd, and against an mcp checkout of supabase/mcp#333 the model's first genuine ClickHouse aggregation succeeds end to end.

Found while running an A/B validation of supabase/mcp#333 through the eval workspace; the run details are in that PR's thread.


packages/platform-lite/src/project/log-seeding.ts
```diff
@@ -90,6 +90,53 @@ CREATE TABLE IF NOT EXISTS storage_logs (
   level text,
   metadata jsonb NOT NULL DEFAULT '{}'::jsonb
 );
+
```
Rodriguespn, line 115: I'm not very familiar with Clickhouse syntax so apologies in advance for the ignortance but where all these sql statement come from?

And what exactly are we "mirroring" here? I would like to avoid having these queries explicit here if there is another place where they are already defined and we can simply import them

packages/platform-lite/src/management-api/debugging.ts
```diff
@@ -72,6 +110,36 @@ export function createDebuggingRoutes(store: ProjectStore): ManagementApiRoutes
...
+ * Anything else surfaces the raw SQL error to the model, which adapts — fine
+ * for exploration, but remember: a query that only works here (postgres-isms)
+ * would FAIL against the hosted ClickHouse endpoint, and vice versa. Extend
+ * only from observed model output, never speculatively.
+ *
+ * KNOWN LIMITATION (time semantics): iso_timestamp_start/end are ignored
+ * (fixed-date seeds), so window-correctness of model queries is NOT exercised
+ * locally. A time-window-discriminating eval needs relative-time seeding.
+ */
+const NUMERIC_LOG_ATTRIBUTES = new Set(['response.status_code', 'status_code', 'execution_time_ms'])
```
Rodriguespn: >  I would like to avoid having these queries explicit here if there is another place where they are already defined and we can simply import them

Same here

Rodriguespn (changes requested):

packages/platform-lite/src/project/log-seeding.ts
barryroodt: Not ignorance at all, it's the right question to ask, and the honest answer deserved to be in the code rather than just here. Added in a0305f8.

Three different origins, none importable unfortunately:

- **The `logs` view shape** mirrors the hosted platform's Logflare/ClickHouse contract for `/analytics/endpoints/logs` (supabase/platform#35096). That schema is platform-internal: no npm package exports it, so the fixture has to model it by hand, same as every other endpoint platform-lite emulates.
- **The `source` names** track mcp's `logsServiceSchema` (`packages/mcp-server-supabase/src/tools/logs.ts`) under the pinned `MCP_SERVER_VERSION`. We could technically import those few strings from the mcp package, but that would couple the fixture to the thing the evals are testing, which felt backwards. Open to it if you'd prefer the coupling over the duplication.
- **The `toInt32OrZero`/`toString` family** reimplements ClickHouse builtins (also no importable artifact); signatures follow the ClickHouse docs, and we only added the ones models genuinely emitted during live runs.

The in-code comments now say all of this, including when to resync. 🙂

packages/platform-lite/src/management-api/debugging.ts
barryroodt: Same story as the seeding file (see the longer reply there), with one extra wrinkle for this spot: the verbatim SQL in `debugging.test.ts` is frozen on purpose. Those are regression fixtures capturing what models actually emitted; if we imported live definitions, the contract tests would follow whatever the code under test does, which defeats them. Provenance note added in a0305f8.

packages/platform-lite/src/project/log-seeding.ts
barryroodt: One correction to my reply above: "none importable" was too strong on one count, and your instinct was worth chasing. The pinned mcp package does export `logsServiceSchema` from its `/platform` entrypoint, and platform-lite already dev-depends on it.

Having dug in, we still shouldn't import it here, for two verified reasons:

- **Wrong namespace**: `logsServiceSchema` enumerates the service preset names (`api`, `edge-function`, `postgres`, ...), which is a different vocabulary from the unified-stream `source` names this view discriminates on (`edge_logs`, `function_edge_logs`, ...). The source names genuinely have no exported artifact; they live inside mcp's preset SQL strings and the `query_logs` tool description.
- **Wrong version**: the resolved devDependency floats within `^0.8.1` (0.8.2 in the current lockfile), while the eval harness runs the `MCP_SERVER_VERSION` pin (0.8.1). Anything we imported would track a version the fixture never exercises.

So the verbatim SQL in the tests stays as the alignment contract: a deliberately frozen copy of what *current* mcp emits (its `getClickHouseLogQuery` presets, plus model-authored queries captured from live runs against a local mcp build). Worth being precise: the 0.8.1 runtime pin doesn't validate that SQL, it predates the ClickHouse endpoint entirely, which is exactly the drift this PR exists to close. The provenance comment records all of this, corrected in 8f71f95. Thanks for pushing on this, the reasoning is better for it 🙂

packages/platform-lite/src/management-api/debugging.ts
```diff
@@ -34,6 +34,49 @@ export function createDebuggingRoutes(store: ProjectStore): ManagementApiRoutes
...
+    // The hosted endpoint is read-only server-side; enforce the same contract on
+    // model-authored SQL. The prefix check only shapes the error message — the
+    // REAL enforcement is the read-only transaction below, which postgres applies
+    // to every statement including data-modifying CTEs. The 400 body carries
+    // `message` because mcp's assertSuccess parses non-2xx bodies as {message}
+    // (the management-API error envelope) — without it the model only sees the
+    // generic "Failed to fetch logs" fallback; `error` kept for shape
+    // consistency with the 200 SQL-error path.
+    const stmt = sql.trim().replace(/;+\s*$/, '')
+    if (stmt.includes(';') || !/^\s*(select|with)\b/i.test(stmt)) {
```
mattrossman: IIUC this PR adds the `logs` view but other tables like `edge_logs`, `postgres_logs`, etc. are still accessible for queries. Does that match how the prod endpoint behaves? Wondering if we need to also restrict these queries to the `logs` view to avoid behavior mismatch in case agent tries to query those other tables directly.

mattrossman:
Surfacing these threads for awareness: [(1)](https://supabase.slack.com/archives/C051L8U2EJF/p1784304358642699?thread_ts=1784278807.909859&cid=C051L8U2EJF) [(2)](https://supabase.slack.com/archives/C08N7894QTG/p1784652501809839?thread_ts=1784624411.491679&cid=C08N7894QTG)

I'm wondering how this approach of translating CH dialect for our PGLite logs backend compares with running actual ClickHouse separate from the PGLite logs backend, using their in-memory [chdb](https://clickhouse.com/chdb) and [chdb-node](https://github.com/chdb-io/chdb-node) bindings? That might get us closer feature parity w/o manual translation for each new syntax the agent tries using, though I'm not sure about the integration lift.

packages/platform-lite/src/management-api/debugging.ts
barryroodt: Good catch, it didn't match. Fixed in c3dfccf: queries on this route now run under a role that can only read the `logs` view, so postgres denies direct table access however it's spelled. If hosted turns out to expose more than the unified stream, we can relax the grant.

barryroodt:
chdb could be a good fit eventually, yeah, thanks for the threads. Today it'd be a second engine (legacy `logs.all` stays on PGlite) and we'd still hand-model the private logs schema, so it removes the smaller half of the work. I'd keep the current two-rewrite translator while it stays this small, and spike chdb if the shim surface starts growing.

packages/platform-lite/src/management-api/debugging.ts
```diff
@@ -84,6 +133,74 @@ export function createDebuggingRoutes(
...
+ * (where source = 'edge_logs').
+ *
+ * PROVENANCE: none of this is importable. The real dialect boundary lives in
+ * the hosted platform's Logflare/ClickHouse backend (supabase/platform#35096,
+ * platform-internal); mcp ships only tool descriptions, and ClickHouse
+ * builtins have no npm artifact. The verbatim SQL in debugging.test.ts is
+ * deliberately FROZEN observed output (regression fixtures) - importing live
+ * definitions would make those contract tests follow the thing they test.
+ *
+ * KNOWN LIMITATION (time semantics): iso_timestamp_start/end are ignored
```
Rodriguespn, line 183: The KNOWN LIMITATION note covers the query params, but the same time-semantics hole is reachable through the SQL itself, where it fails silently instead of loudly.

For example, seeds carry fixed past dates while `now()` is real wall-clock, so `select count(*) from logs where source = 'edge_logs' and timestamp > now() - interval '24 hours'` returns `200 {result: []}` with no error, and the agent concludes "no errors occurred". 

Until relative-time seeding lands, a guard that rejects `now()`/`current_timestamp` in logs SQL (loud, like the other guards) would keep evals honest.

packages/platform-lite/src/project/log-seeding.ts
```diff
@@ -259,6 +361,26 @@ async function seedAuthLog(
...
+      log.source,
+      log.level,
+      log.metadataJson,
+    ]
+  );
+}
+
 function metadataText(
   metadata: Record<string, unknown>,
   keys: string[]
```
Rodriguespn, line 476: This branch relabels `function_edge_logs` rows as `function_logs`, so one seeded `edge-function` row appears 3× in the unified view (unfiltered counts and `group by source` overcount 3× and show a phantom console-log stream). 

I suggest treating `function_logs` as unmodeled (loud error, like `workflow_run_logs`) until it can be seeded independently, and pinning `group by source` output in a test.

packages/platform-lite/src/management-api/debugging.ts
```diff
@@ -84,6 +133,74 @@ export function createDebuggingRoutes(
...
+  }
+  const physical = PHYSICAL_RELATIONS.exec(sql);
+  if (physical) {
+    throw new Error(
+      `relation '${physical[1]}' is not queryable on this endpoint — query the unified 'logs' stream and filter with where source = '${physical[1]}'`
+    );
+  }
+  return sql
+    .replace(
+      /\blog_attributes\['([^']+)'\]/gi,
```
Rodriguespn: Missing-key semantics diverge from hosted: [ClickHouse Map access returns `''` for absent keys](https://clickhouse.com/docs/en/sql-reference/data-types/map), `->>` returns NULL, so `log_attributes['error'] = ''` silently matches nothing locally. 

Compile to `coalesce(log_attributes->>'${key}', '')`. Also make the regex whitespace-tolerant (`\[\s*'([^']+)'\s*\]`): `log_attributes[ 'k' ]` currently falls through to jsonb subscripting with quote-wrapped values.

packages/platform-lite/src/management-api/debugging.ts
```diff
@@ -41,6 +41,55 @@ export function createDebuggingRoutes(
...
+    // The hosted endpoint is read-only server-side; enforce the same contract on
+    // model-authored SQL. The prefix check only shapes the error message — the
+    // REAL enforcement is the read-only transaction below, which postgres applies
+    // to every statement including data-modifying CTEs. The 400 body carries
+    // `message` because mcp's assertSuccess parses non-2xx bodies as {message}
+    // (the management-API error envelope) — without it the model only sees the
+    // generic "Failed to fetch logs" fallback; `error` kept for shape
+    // consistency with the 200 SQL-error path.
+    const stmt = sql.trim().replace(/;+\s*$/, '');
+    if (stmt.includes(';') || !/^\s*(select|with)\b/i.test(stmt)) {
```
Rodriguespn: Suggestion: This gate 400s SQL hosted accepts: leading comments (`-- …\nselect`), semicolons inside string literals (`like '%;%'`), and `(select …)`. 

The role + read-only transaction do the real enforcement. Please strip leading comments before the prefix test and ignore `;` in quoted literals, so agents aren't failed for a fixture artifact.

packages/platform-lite/test/clickhouse-logs.test.ts
```diff
@@ -0,0 +1,364 @@
...
+    }
+  );
+});
+
+// Route-level contract: the read-only guarantee must hold at the HTTP boundary
+// where model-authored SQL arrives, not just in the translator. A real
+// ProjectInstance (constructor is init-free; the route only touches logsDb) in
+// a real Map — the exact ProjectStore shape, no casts. Dedicated instance so a
+// failing guard can't poison the other tests' fixture rows.
+describe('/v1/projects/:ref/analytics/endpoints/logs route', () => {
```
Rodriguespn, line 400: The ignored-timestamp behavior is documented but not test-pinned. A future "fix" that starts filtering would pass this suite while emptying every seeded scenario. 

Please add a route test passing a now-ish `iso_timestamp_start/end` and asserting seed rows still return. 

Also worth mentioning that a modeled-but-unseeded source returning `200 {result: []}` with no `error`, and any execution of `toInt64OrZero`/`toUInt32OrZero` (note `v::numeric` gives `'10.5'` → `10.5` where [ClickHouse returns `0`](https://clickhouse.com/docs/en/sql-reference/functions/type-conversion-functions#toint32orzero)) is missing.

Rodriguespn:
Thanks again for tackling this @barryroodt. The PR is in a great shape. Included some comments but overall is good

packages/platform-lite/src/management-api/debugging.ts
barryroodt: Confirmed, `timestamp > now() - interval '24 hours'` returns `200 {result: []}`. Wall-clock functions now reject loudly (7d84ccf), but only when the statement reads `logs`, so the `now()` orientation probes your #333 A/B saw still pass (d0c6b81).

packages/platform-lite/src/project/log-seeding.ts
barryroodt: You're right, and it's 3× exactly. I started with your suggestion, then found mcp 0.9.0 has an `edge-function-runtime` preset reading `source = 'function_logs'`, so rejecting it would break on the next pin bump. It gets its own table and seed source instead, so an `edge-function` seed no longer writes there. Fixed in 7d84ccf.

packages/platform-lite/src/management-api/debugging.ts
barryroodt: Both correct, fixed in 7d84ccf. Map access compiles to `coalesce(log_attributes->>'k', '')` and the subscript is whitespace-tolerant. One thing worth knowing: the spaced form was throwing `invalid input syntax for type json`, not just returning quote-wrapped values.

packages/platform-lite/src/management-api/debugging.ts
barryroodt: Agreed, all four 400'd. The gate now tests a copy with literals and comments blanked, so those cases pass while a genuine second statement still doesn't (7d84ccf). Did you get an answer on whether hosted hard-rejects non-SELECT? Our 400 path should match whatever it does.

packages/platform-lite/test/clickhouse-logs.test.ts
barryroodt: All three added in 7d84ccf. Your `toInt64OrZero` note applied to the whole family, they parse a whole integer or return 0 now. One correction worth having: out-of-range strings return 0 rather than wrapping, so `toInt32OrZero('2147483648')` is 0. The wrapping in the docs is the numeric overload. Checked on play.clickhouse.com.

Rodriguespn (approved):
LGTM. Nice work! 

Approving this to unblock you, but I would suggest porting the `logs_reader`/read-only wrapper to the legacy `logs.all` route as a fast follow-up. 
It still runs agent SQL with write privileges, so an agent on the older path can mutate log state that the new hardened route then faithfully serves. The guard machinery already exists a few lines below, so it should be cheap to reuse.

gregnr:
From what I can piece together, we're using PGlite to simulate the DB behind the CH endpoint? Since we want our evals to be as accurate as possible, I'm worried that seeding/simulating logs will be doing us a disservice. Have we considered [chdb-wasm](https://www.npmjs.com/package/chdb-wasm) or [chdb-node](https://github.com/chdb-io/chdb-node) instead? Beyond ClickHouse, how are application logs surfaced right now (or is it all seeded/fake data)?

barryroodt:
> From what I can piece together, we're using PGlite to simulate the DB behind the CH endpoint? Since we want our evals to be as accurate as possible, I'm worried that seeding/simulating logs will be doing us a disservice. Have we considered [chdb-wasm](https://www.npmjs.com/package/chdb-wasm) or [chdb-node](https://github.com/chdb-io/chdb-node) instead? Beyond ClickHouse, how are application logs surfaced right now (or is it all seeded/fake data)?

Thanks @gregnr, you're correct yes: PGlite behind a small ClickHouse-dialect translator.

Your last question first: it's all seeded. Seven evals ship a `remote/logs.jsonl`, every timestamp absolute, and nothing ingests real logs.

I spiked both bindings (`chdb-node`& `chdb-wasm`) against the SQL `mcp-server-supabase@0.9.0` actually emits (current pinned version of 0.8.1 does not emit ClickHouse queries). `chdb-wasm` is the better of the two and its dialect fidelity is genuinely better than ours. It wants Node >=23 though, and we're pinned at 22.

This PR is essentially a "forward-compatibility" update for when MCP#333 (`query_logs`) is accepted and deployed.
It stays a no-op until we bump the mcp package version. No eval scores will be updated.

I don't see any harm in merging this now as is, then in a follow-up PR, bump the mcp & node versions + replace the dialect translator with `chdb-wasm` for better fidelity.

Rodriguespn:
> I don't see any harm in merging this now as is, then in a follow-up PR, bump the mcp & node versions + replace the dialect translator with chdb-wasm for better fidelity.

Initially, this was my thought because "This PR is essentially a "forward-compatibility" update for when MCP#333 (query_logs) is accepted and deployed" but then I remembered that pkg.pr.new can point to the latest commit of a PR, if we use the PR number instead of the commit hash:

```bash
https://pkg.pr.new/@supabase/mcp-server-supabase@333
```

What if we pointed the MCP server in our [package.json](https://github.com/supabase/evals/blob/3672889f714cd7db96b22d419d449d4ef1ec5f4d/packages/platform-lite/package.json#L31) to this release version and once it's merged, we point to the new release?

Wdyt @barryroodt?

---

Note: "The PR-number URL always points to the latest PR commit" isn't documented, but the code confirms it: publishes advance a per-ref cursor to the new commit ([`publish.post.ts`](https://github.com/stackblitz-labs/pkg.pr.new/blob/main/packages/app/server/routes/publish.post.ts)), and non-SHA refs resolve through that cursor ([`tarball-resolver.ts`](https://github.com/stackblitz-labs/pkg.pr.new/blob/main/packages/app/server/middleware/tarball-resolver.ts)).

gregnr:
Makes sense - PGlite translation was already in there, we're just adding CH support. Merging as-is SGTM with plans to swap in chdb-wasm after

Rodriguespn:
Did one last review pass, nothing to flag other than what Greg pointed out about the follow up (cc @barryroodt). Shipping it 🚢

PR #100, opened 2026-07-21
feat(eval): query_logs tool selection on hosted-shaped platform

jordienr:
## What

One MCP-mode eval verifying the agent successfully uses `query_logs` to investigate edge function logs on a hosted-shaped platform-lite fixture:

- `investigate-logs-002-query-logs-tool-selection` — asserts `query_logs` was called (and `get_logs` was not), scored deterministically on `ctx.toolCalls[].endpoint`.

## Why

Review feedback on supabase/mcp#333 (Rodriguespn): "curious if the agent will correctly call `query_logs` when working with the hosted version ... we should have an eval scenario." That PR now hides `get_logs` from `tools/list` whenever the platform implements `queryLogs` (registration-driven, not description-driven), so this exercises `query_logs` end-to-end on the fixture where it's the only logs tool listed.

## Dropped: the self-hosted arm

The original version of this PR also had `investigate-logs-003-tool-selection-self-hosted`, asserting the agent uses `get_logs` (not `query_logs`) on a self-hosted-shaped platform. That scenario is **untestable through this harness** and has been removed:

`supabaseMcpServer()` (`packages/core/src/index.ts`) always spawns `npx @supabase/mcp-server-supabase@<version> --api-url <backend>`, which always constructs its platform via `createSupabaseApiPlatform()` (`packages/mcp-server-supabase/src/platform/api-platform.ts`). That implementation defines `queryLogs` **unconditionally** — it doesn't vary by `--api-url`, including when pointed at platform-lite. The actual self-hosted/CLI behavior (`get_logs`-only) comes from a *different*, external `SupabasePlatform` implementation (the CLI's own adapter) that this harness never spawns. There's no way to make the spawned server lack `queryLogs` from evals today.

That registration behavior is already covered by `supabase/mcp`'s own unit tests (`server.test.ts`: "debugging tools show get_logs when the platform does not implement query_logs"; `debugging-tools.test.ts`). If we want a real evals-level self-hosted check in the future, `@supabase/mcp-server-supabase` would need some way to force a `queryLogs`-less platform for testing (e.g. a flag or a local-stack-shaped runtime), which is out of scope here.

## Still blocked — not runnable yet

`query_logs` only exists in **unmerged** supabase/mcp#333. `MCP_SERVER_VERSION` here is pinned to a version without it, so this eval can't pass until:
1. supabase/mcp#333 merges and a new `@supabase/mcp-server-supabase` version ships `query_logs`, and
2. `MCP_SERVER_VERSION` is bumped to that version in `packages/core/src/index.ts`.

Validated with `pnpm eval:dry` (loads, parses, plans as `mode=tools runtime=platform-lite`). Not executed — needs the version bump above.


Rodriguespn:
Closing as https://github.com/supabase/mcp/pull/333 was merged and these evals were just to test the change

PR #101, opened 2026-07-21
feat: upload benchmark results to Braintrust

Rodriguespn:
Uploading benchmark results to Braintrust.

This is an additive dual-write agreed in [Slack](https://supabase.slack.com/archives/C0BAGJBL49E/p1784579768400889): `eval-results.json` stays exactly as it is (verified byte-identical) and remains the source of truth; Braintrust becomes a second destination for exploring and comparing results. Design doc: [Braintrust as the storage layer of Supabase evals](https://app.notion.com/p/supabase/Braintrust-as-the-storage-layer-of-Supabase-evals-3a35004b775f815bbffbf91505747156).

## How our results map to Braintrust

A **Braintrust experiment** is a group of agent runs against a set of scenarios. We create one experiment per model on every refresh, named `<experiment>@<git-sha>` (e.g. `claude-code-sonnet-5@fe0e167`), so refreshes stay comparable over time — each new experiment is automatically diffed against the previous one for the same model.

Inside an experiment, **each eval is one row**. The row's input is the scenario prompt (what Braintrust uses to match the same scenario across models), the output is the agent's final report, and the scenario dimensions (stage, product, topic, interface…) are attached as metadata and tags for filtering.

Example of a trace: [claude-code-sonnet-5@59fc9c7-raw-transcript](https://www.braintrust.dev/app/supabase.io/p/(Pedro)%20Running%20agent%20skills%20evals%20through%20github%20actions%20job/experiments/claude-code-sonnet-5%4059fc9c7-raw-transcript)

Opening a row shows the **full trace** of what the agent did:

- a **turn** groups everything the agent did between user prompts (evals send a single prompt, so one turn is the norm);
- inside it, each **step** is one model API call, carrying the model's verbatim text/thinking and that call's own token counts — you can watch the context grow step by step;
- each **tool call** hangs under the step that requested it, with its args and result; subagent work (Claude Code's `Task`) nests under the tool call that spawned it.

The shape is identical for Claude Code, Codex and ai-sdk: each agent's parser knows its own transcript format, and a shared agent-agnostic model (`AgentTrace` in core) assembles the tree.

- **Scoring** appears as a single `passed` scorer — matching how the web UI headlines pass/fail — with the full checks list inside its output instead of one scorer per check, so the experiment table keeps one clean score column. When a check used an LLM judge, the judge call is nested under the scorer with the rubric, the exact input the judge read, and its model/duration/tokens: you audit what the judge saw instead of trusting its notes.

- **Token metrics** follow Braintrust's conventions (`prompt_tokens`, `prompt_cached_tokens`, …), taken from each agent's own accounting; Claude input tokens are normalized to include cache reads/writes so counts compare across agents. The row covers the real run wall-clock; span ordering inside is synthetic (CLI transcripts carry no per-event timestamps) and flagged `stepTiming: synthetic-order-only`.

Every row also carries two **attachments** for forensics: the raw result JSON and the agent CLI's verbatim transcript (`<evalId>.transcript.jsonl`).

image (1)

## Experiment metadata

Each experiment carries exactly the metadata fields from the design doc's table — the decomposed config identity (`agent`, `modelProvider`, `modelId`, `reasoningEffort`, `skills`) plus the run context (`branch`, `prNumber`, `experimentSuite`, `trigger`, `runId`, and `latestMain` for the current benchmark baseline) — so each is independently filterable/groupable in the Braintrust UI. `skills` is stored as a stable sorted comma-joined string, the same shape the baseline lookup queries on.

## Baseline comparison

A full benchmark refresh on `main` stamps its experiments `latestMain` (old holders are demoted only after the whole batch uploads, so a crashed refresh never loses the baseline). PR runs look up the baseline for the same config identity and set it as the Braintrust base experiment, so every PR experiment opens pre-diffed against the latest main results.

## In CI

After the export step, a `continue-on-error` upload runs for benchmark evals (missing credentials = a warning, never a failed job). The workflow upserts a single PR comment showing each experiment's pass rate, the Δ vs the latest-main baseline (linking to the pre-compared view), and pre-filtered links to this PR's experiments and the latest main results — see [the comment on this PR](https://github.com/supabase/evals/pull/101#issuecomment-5049354811) for a live example produced by a simulated run. Ad hoc custom-suite dispatches upload under a branch-suffixed name so they're never mistaken for tracked results. Result files from before these fields existed still upload via a flat fallback.

## How to review

Core logic, in dependency order:

1. `packages/core/src/transcript/agent-trace.ts` + `packages/core/src/agents/*` — parsing CLI transcripts into the shared `AgentTrace` model and per-call usage capture.
2. `packages/core/src/braintrust-spans.ts` — mapping a result + trace into the Braintrust span tree, row metadata, and token metrics (unit-tested).
3. `apps/framework/scripts/upload-braintrust.ts` — the uploader: experiment naming/metadata, baseline lookup/rotation, summary markdown for the PR comment.
4. `apps/framework/lib/result-files.ts` — result-file scan shared with the JSON export so the two destinations can't disagree.
5. `.github/workflows/eval-refresh.yml` — the upload + PR-comment steps (both `continue-on-error`).

Mechanical/generated (skim only): `apps/web/src/data/eval-results.json` and `regression-eval-results.json` are regenerated eval artifacts (the big line count is churn from a results refresh, verified byte-identical export behavior); `pnpm-lock.yaml` is the `braintrust` dependency.

Closes [AI-907](https://linear.app/supabase/issue/AI-907/upload-results-to-braintrust-and-validate-experiments)


.github/workflows/eval-refresh.yml
```diff
@@ -209,37 +228,57 @@ jobs:
...
-                continue
-              fi
-
-              if [ -n "$experiments_override" ]; then
-                experiments_json="$(jq -Rc 'split(",") | map(gsub("^\\s+|\\s+$"; "")) | map(select(length > 0))' <<< "$experiments_override")"
-              else
-                experiments_json="$(pnpm --silent eval -- list --experiment-suite "$experiment_suite")"
-              fi
 
+          if jq -e 'index("custom") != null' <<< "$suite_json" > /dev/null 2>&1; then
```
Rodriguespn, line 232: To run one (or more) arbitrary pair evals/experiments (ex: investigate-security-001-public-table/codex-gpt5.4-mini) without being tied to a pre defined batch (benchmark or regression). This workflow only run in branhces. It fails if we try to run on `main`

Rodriguespn:
### Braintrust benchmark results

| Experiment | passed | Δ vs main |
|---|---|---|
| [claude-code-sonnet-5](https://www.braintrust.dev/app/supabase.io/p/(Pedro)%20Running%20agent%20skills%20evals%20through%20github%20actions%20job/experiments/claude-code-sonnet-5%401234abc-pr-sim5) | 3/3 · 100% | [⚪ 0% vs main@aaaaaaa-f5d4a576](https://www.braintrust.dev/app/supabase.io/p/(Pedro)%20Running%20agent%20skills%20evals%20through%20github%20actions%20job/experiments/claude-code-sonnet-5%401234abc-pr-sim5?c=claude-code-sonnet-5%40aaaaaaa-f5d4a576) |

[Braintrust results for PR #101](https://www.braintrust.dev/app/supabase.io/p/(Pedro)%20Running%20agent%20skills%20evals%20through%20github%20actions%20job/experiments?search=%7B%22filter%22%3A%5B%22metadata.prNumber%2520%253D%2520'101'%22%5D%7D) · [Latest main results](https://www.braintrust.dev/app/supabase.io/p/(Pedro)%20Running%20agent%20skills%20evals%20through%20github%20actions%20job/experiments?search=%7B%22filter%22%3A%5B%22metadata.latestMain%2520%253D%2520true%22%5D%7D)

PR #103, opened 2026-07-22
fix: retry sandbox image build on transient registry errors

Rodriguespn:
Currently, eval CI jobs fail whenever Docker Hub has a blip: during [this run](https://github.com/supabase/evals/actions/runs/29926252232) Docker Hub returned 502 and 7 jobs failed. e.g. [this job](https://github.com/supabase/evals/actions/runs/29926252232/job/88944704724):

This PR adds a retry with backoff system (5s/30s/60s) that should prevent run failures caused by transient Docker Hub outages/rate limits. Every build failure is retried — the registry-error strings come from BuildKit/containerd internals with no stable taxonomy to classify against, so a deterministic failure just wastes the ~95s schedule before failing.


mattrossman (approved):
Surprised there isn't a native feature for this in Docker / GH actions, so this looks like a reasonable change!

PR #105, opened 2026-07-22
Update README.md with correct clone

czenko:
The clone command didn't use the correct SSH name.


mattrossman (approved):

mattrossman:
Thanks for catching that, the folder structure I use on my local clone must have slipped in there!

PR #108, opened 2026-07-22
feat: add docs homepage quickstart eval

czenko:
Closes [DOCS-1199](https://linear.app/supabase/issue/DOCS-1199/add-one-happy-path-quickstart-eval-test)

## Problem
The docs homepage shows a "Help me get set up with Supabase" prompt. That flow installs the Supabase CLI and Supabase plugin, detects or initializes the project, and suggests next steps, but no eval covers it. CI wouldn't catch a regression in the CLI or plugin install steps, and the sandbox pre-satisfying those steps would mask a failure even in a new eval.

## Solution
- Add `evals/build-cli-004-quickstart-nextjs`, seeded with a bare Next.js app, to exercise the docs homepage quickstart prompt end-to-end.
- Add per-eval `skills: []` and `skipCliInstall: true` overrides so the eval tests the CLI and plugin install steps directly, instead of relying on the sandbox to pre-satisfy them.
- Add an experiment-level `skipEval` predicate so the no-skills experiment skips evals whose skills list is already empty, avoiding a duplicate run.
- Fix the eval-refresh workflow's pair discovery to respect `skipEval`. Without this fix, `publish-results` tried to download an artifact for a pair that `skipEval` prevents from ever being produced. This broke CI on the first run of this PR.

## Manual testing
- Run `pnpm eval:dry -- --eval build-cli-004-quickstart-nextjs --experiment claude-code-sonnet-5,claude-code-sonnet-5-no-skills`. Only `claude-code-sonnet-5` should be planned. The no-skills experiment should print `SKIP ... (skipEval)`.
- Check `evals/build-cli-004-quickstart-nextjs/EVAL.ts` for the checks that confirm the CLI and plugin were installed, rather than already present. That's the part most likely to break if the sandbox's base image changes.
- Add the `run-evals-changed` label to confirm `publish-results` completes without the artifact-download failure.


nrichers (commented):
@czenko as mentioned on Slack, I think you just broke the ice on getting docs into Evals! 👏  

I tested your PR locally and for the intent of adding a happy path smoke test, LGTM. 👍 I will defer to the Evals team for approval of the actual code changes as these are outside of my wheelhouse. 

This is a great first step towards our a AEO testing aspirations. Way cool and thank you so much! 🚀

--- 
**Verdict:** Approve with a note

| Check | Result |
|---|---|
| `pnpm --filter @supabase-evals/framework typecheck` | ✅ pass |
| `pnpm --filter @supabase-evals/core test` | ✅ 47/47 pass |
| `pnpm --filter @supabase-evals/sandbox test` | ✅ 35/35 pass |
| `SANDBOX_DOCKER_TESTS=1 ... test:docker -t skipCliInstall` | ✅ pass — CLI genuinely absent, shim doesn't crash setup |
| `pnpm eval:dry --eval build-cli-004-quickstart-nextjs --experiment claude-code-sonnet-5,claude-code-sonnet-5-no-skills` | ✅ matches PR claim exactly: `SKIP claude-code-sonnet-5-no-skills ... (skipEval)`, `PLAN claude-code-sonnet-5 ...` |
| CI | Only Vercel preview checks run on this repo; no required checks configured |

**Verified:**
- `skills`/`skipCliInstall` frontmatter plumbing is threaded correctly end-to-end: `eval-metadata.ts` → `run-eval.ts` (`ev.metadata.skills ?? exp.skills`) → `agent-environment.ts` → `local-stack-runtime.ts`/`supabase.ts` (skip both CLI install and the CLI wrapper when `skipCliInstall`).
- `skipEval` on `claude-code-sonnet-5-no-skills` and the new `list --eval` filtering in `run-eval.ts`/`eval-refresh.yml` correctly avoid planning a redundant duplicate run — confirmed live via `eval:dry`.
- New eval's scorer (`EVAL.ts`) checks are sound: `supabase/config.toml` existence, `supabase --version` runnable, an LLM-judge rubric for "relevant next steps," and a regex check for the plugin-install command actually being attempted (explicitly *not* asserting install success, with a comment explaining why — reasonable scoping).

**Notes (non-blocking):**
- The checked-in `apps/web/src/data/regression-eval-results.json` currently contains a stale entry: `claude-code-sonnet-5-no-skills` × `build-cli-004-quickstart-nextjs` shows `passed: true`, but per this PR's own `skipEval` logic, that experiment should no longer run against this eval at all. Tracing the two `chore: refresh eval results` bot commits: the first (`ab5421a`, 2026-07-22 23:22 UTC) ran *before* the `skipEval` fix (`cc8b18a`, 2026-07-23 19:18 UTC) and recorded that result; the second (`f93b41a`, 19:24 UTC) ran after the fix but used the PR-triggered `run-evals-changed` path, which calls `export-results --merge`. `export-results.ts`'s merge logic (`apps/framework/scripts/export-results.ts:274-289`) only overwrites `experiment::eval` keys present in the *new* results — it never prunes keys that stop being produced. Since the skipped combo now produces no artifact at all, the stale pre-fix "passed" result will persist through every future merge-mode refresh, only self-correcting on the next full (non-merge) scheduled regression run. This is a pre-existing gap in `export-results.ts`'s merge semantics, not something this PR broke, but this PR is the first to exercise `skipEval` and is shipping a visibly-stale row into `main` as a result. Worth a follow-up (either prune-on-merge in `export-results.ts`, or manually drop that row before merging).
- `claude-code-sonnet-5`'s own real (non-dry) regression result for the new eval shows `passed: false` (`supabase CLI is installed and runnable` failed — CLI not found after the agent's `npm install -g supabase`), while everything else passed (init, plugin-install attempt, next-steps judge). Worth confirming with the author whether this is expected/tracked behavior (a real capability gap the eval is meant to surface) or a PATH-related environment artifact — not a reason to block the PR either way, since the eval mechanically does exactly what it's designed to do.

czenko:
@nrichers Thanks for flagging the merge-mode gap. Fixes in [`claude/nervous-allen-83c06e`](https://github.com/supabase/evals/commit/3fbc751ed5e7abf2ec3ec5452fc2dddfb73bcf6c).

`export-results.ts` now accepts a `--requested experiment::eval,...` flag; merge mode drops an existing row when its key was requested this run but produced no result file , while leaving rows untouched for pairs that simply weren't part of this run.

evals/build-cli-004-quickstart-nextjs/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This guide explains how to add an eval for another AI setup prompt from the docs, such as a different framework's quickstart page.
+
+1. Create a sibling folder, for example `build-cli-005-quickstart-<framework>`.
+2. Copy the target prompt verbatim into `PROMPT.md`. Don't paraphrase it or add detail that the prompt doesn't already give. Keep `skills: []` and `skipCliInstall: true` in the frontmatter if the prompt has the agent install its own tooling.
+3. Seed `local/` with a minimal, unmodified starter for that framework. Don't include a `supabase/` directory, so the "is Supabase already initialized" step has something real to check.
+4. Reuse this eval's `EVAL.ts` checks. Change only the judge rubric's framework-specific wording.
+5. Validate the eval before you open a PR. Run `pnpm eval:dry -- --eval <new-eval-id> --experiment claude-code-sonnet-5` to check the setup, then run `pnpm eval -- --eval <new-eval-id> --experiment claude-code-sonnet-5` for a full pass.
+
+For the full eval-authoring workflow, including frontmatter fields, suite selection, and submitting for review, see `CONTRIBUTING.md`.
```
czenko:
```suggestion
# Adding an eval for another docs guide

This guide explains how to add an eval for another AI setup prompt from the docs. The prompt can come from a quickstart page for a different framework, or from any other docs guide that gives an agent a setup prompt to follow.

1. Create a sibling folder. Name it after the guide you're adding, for example `build-cli-005-quickstart-<framework>`.
2. Copy the target prompt verbatim into `PROMPT.md`. Don't paraphrase it or add detail that the prompt doesn't already give. Keep `skills: []` and `skipCliInstall: true` in the frontmatter if the prompt has the agent install its own tooling.
3. Seed `local/` with a minimal, unmodified starter that matches the guide's context. Don't include a `supabase/` directory, so the "is Supabase already initialized" step has something real to check.
4. Reuse this eval's `EVAL.ts` checks as a starting point. Adjust the checks to match what the new guide's prompt requires, including the judge rubric's wording.
5. Validate the eval before you open a PR. Run `pnpm eval:dry -- --eval <new-eval-id> --experiment claude-code-sonnet-5` to check the setup, and then run `pnpm eval -- --eval <new-eval-id> --experiment claude-code-sonnet-5` for a full pass.

For the full eval-authoring workflow, including frontmatter fields, suite selection, and submitting for review, see `CONTRIBUTING.md`.
```

nrichers (approved):
LGTM, thank you for addressing my comment. 🚀

mattrossman (changes requested):
Hey @czenko I tested running this new eval locally. During that local run, the agent failed to install `supabase` CLI with `npm` because the Node image installs node as `root` and doesn't expose `sudo` access, so it gets stuck on the `npm install -g supabase` step:

```bash
$ npm install -g supabase 2>&1

Exit code 243
npm error code EACCES
npm error syscall mkdir
npm error path /usr/local/lib/node_modules/supabase
npm error errno -13
npm error Error: EACCES: permission denied, mkdir '/usr/local/lib/node_modules/supabase'
...
npm error It is likely you do not have the permissions to access this file as the current user
npm error
npm error If you believe this might be a permissions issue, please double-check the
npm error permissions of the file and its containing directories, or try running
npm error the command again as root/Administrator.
```

I believe the CI run failed for the same reason:

https://github.com/supabase/evals/pull/108/changes#diff-3b6aa05c44031b2fe9a909f86aa07cf051bd13e1ee6fb18c71d159b5f7383ce6R29-R31

I figure a typical user's machine will have Node configured in a way that `npm install -g` "just works" for a user without needing `sudo`, maybe we should update the Dockerfile to configure it this way to make the scenario more fair?

I found this [Global npm dependencies](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#global-npm-dependencies) section in the docs for the official Node.js docker images, which recommends the following:

```dockerfile
ENV NPM_CONFIG_PREFIX=/home/node/.npm-global

ENV PATH=$PATH:/home/node/.npm-global/bin # optionally if you want to run npm global bin without specifying path
```

The `ENV PATH` line may need to be adapted for our use case, because we override `PATH` with an explicit `SANDBOX_PATH` list [here](https://github.com/supabase/evals/blob/f22c7c4927c1bb3dc883b67030bf6270f2d9875f/packages/sandbox/src/docker-sandbox.ts#L181), so we can instead add it like:

```diff
--- a/packages/sandbox/src/docker-sandbox.ts
+++ b/packages/sandbox/src/docker-sandbox.ts
@@
 const SANDBOX_PATH = [
+  "/home/node/.npm-global/bin",
   "/usr/local/sbin",
   "/usr/local/bin",
   "/usr/sbin",
   "/usr/bin",
   "/sbin",
   "/bin",
 ].join(":");
 ````

I think we'll also need to make sure the `/home/node/.npm-global` folder isn't created as `root` during image build, otherwise subsequent `npm install -g` may not work. Note the `npm install -g skills` line implicitly initializes this folder as whichever `user is active (`root` unless otherwise specified in the Dockerfile), so we could explicitly create the folder + set up the npm config prefix beforehand.

```diff
--- a/packages/sandbox/Dockerfile
+++ b/packages/sandbox/Dockerfile
@@
 FROM node:22-slim

 RUN apt-get update -qq && apt-get install -y -qq --no-install-recommends \
     ca-certificates curl git docker.io postgresql-client \
   && rm -rf /var/lib/apt/lists/*

+USER node
+RUN mkdir -p /home/node/.npm-global
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+
 # Vercel's agent-skills CLI, used to install agent skills from local sources
 # into each session's workspace (both modes). Baked in (pinned) so installs are
 # offline and instant rather than an npx download per session.
 ARG SKILLS_CLI_VERSION
 RUN npm install -g "skills@${SKILLS_CLI_VERSION}"
```
I tested this change locally and verified it allows the new eval to pass 4/4 checks, vs failing one of the checks otherwise.

mattrossman:
Also may want to add a test to lock in the behavior:

```ts
it(
  "npm install -g lands on PATH without sudo/workarounds",
  { timeout: TEST_TIMEOUT_MS },
  async () => {
    const image = await ensureSupabaseSandboxImage();
    const sandbox = await DockerSandbox.create({ image, network: "host" });
    try {
      const install = await sandbox.runShell("npm install -g supabase 2>&1");
      expect(install.ok).toBe(true);

      const version = await sandbox.runShell("supabase --version");
      expect(version.ok).toBe(true);
    } finally {
      await sandbox.stop();
    }
  },
);
```

czenko:
@mattrossman Thanks for digging into this. Thoughtful, thorough feedback. 

Applied basically as you proposed:

- `packages/sandbox/Dockerfile`: switch to USER node, set NPM_CONFIG_PREFIX=/home/node/.npm-global before installing the skills CLI, so it and the agent's own npm install -g supabase both land in a node-owned prefix.
- Added /home/node/.npm-global/bin to SANDBOX_PATH.
- Also added your suggested regression test
- Cleaned up commit history

evals/build-cli-004-quickstart-nextjs/EVAL.ts
```diff
@@ -0,0 +1,107 @@
...
+  judge,
+  serializeTranscript,
+  type CheckResult,
+  type LocalStackEvalContext,
+  type LocalStackScorer,
+} from '@supabase-evals/core';
+import { stripIndent } from 'common-tags';
+
+const PLUGIN_INSTALL_PATTERN =
+  /npx\s+plugins\s+add\s+supabase-community\/supabase-plugin/i;
```
mattrossman: I experienced a local run that failed because the agent tried to run:
```bash
npx --yes plugins add supabase-community/supabase-plugin
```
The regex didn't match because of the `--yes`, perhaps it should be more forgiving?

Image

czenko: Thanks. Will look into this on Monday.

czenko: Pushing a fix for the regex, @mattrossman.
As another note, if there is more feedback that seems to be low-hanging, feel free to commit it directly so this can be merged. Or maybe we can review it together live? I admit that a lot of this PR is AI-assisted and your context and expertise will be valuable for me.

packages/sandbox/src/local-stack-runtime.ts
```diff
@@ -98,17 +96,21 @@ export function localStackRuntime(
...
 
       const mcpServers = await resolveMcpServers(options, hosted);
 
-      const baseAddendum =
-        'The Supabase CLI (`supabase`), docker, psql, git, and curl are installed in the workspace. ' +
-        'Use the bash tool to run commands (the working directory is always the workspace root) ' +
-        'and the files tools to inspect and modify files. ' +
-        'Services started with `supabase start` are reachable on their default 127.0.0.1 ports.';
+      const baseAddendum = skipCliInstall
+        ? 'docker, psql, git, and curl are installed in the workspace, but the Supabase CLI is not — install it yourself. ' +
```
mattrossman: In a follow-up PR I'd like to revisit if we need this `baseAddendum` at all, it seems like it'd give agents an unfair head start on stuff that it should be able to discover on its own.

For now, I'm going to at trim the "install it yourself" part of this diff to avoid giving any explicit instructions beyond what's in the eval's prompt.

mattrossman: Removed some of the hints in https://github.com/supabase/evals/pull/108/changes/4420aba4b66672b1f1a2aefc632b0552df7c6d7d and reduced some of the templating duplication

mattrossman (approved):

PR #118, opened 2026-07-24
Fold the eval-source workspace into evals (M2)

barryroodt:
Stacked on #109 (retarget to main when it merges). Moves the [eval-workspace](https://github.com/supabase/eval-workspace) glue into this repo, per the direction agreed in the workspace-layout Slack thread: the sources under test (docs, skills, MCP server) now live beside the harness that tests them.

## What's here (commit per step)

1. **contentApiUrl threading** (`packages/core`): `SUPABASE_CONTENT_API_URL` / a `contentApiUrl` option adds `--content-api-url` to the server args, pointing `search_docs` at an alternative docs index. Requires a server that understands the flag (the `SUPABASE_MCP_SERVER_PATH` build; upstream PR: [mcp#343](https://github.com/supabase/mcp/pull/343)); inert otherwise. Retires the last evals enabler patch.
2. **Verbatim copy** of the glue under `workspace/` + the `mise` task file at the root (Supabase convention) — the reviewable copy boundary, no adaptation.
3. **Re-root**: evals is the host now, not a managed clone. `evals/` path prefixes drop, `cd evals && pnpm eval` becomes a root `pnpm eval`, the evals manifest entry + marker plumbing die, and the mcp clone is replaced by the pinned `submodules/mcp` working tree (enabler patches apply there as marker commits; `submodule."submodules/mcp".ignore=all` keeps patch dirt out of host status, so pin bumps are a deliberate `git add`). The supabase docs monorepo stays an opt-in sparse clone.
4. **Docs + CI**: `workspace/README.md` is the ops doc (bootstrap, the three edit→eval loops, head-to-head A/B, provenance receipts); a free self-test workflow (shell + node stdlib, no keys/docker/model spend) runs on `workspace/**` PRs.

## Verification

- Self-tests from the repo root: status.test **19/0**, hooks.test **8/0** (real pre-push-guard lifecycle against the initialized mcp submodule), ab.test **22/0**.
- Core suite (incl. 2 new contentApiUrl tests): **58 passed**; framework typecheck clean.
- **Fresh-consumer smoke**: plain `git clone` → `status` prints the exact bootstrap steps → `setup.sh` reaches Ready in ~16s (install, selective submodule init, patch markers, keychain keys) → all three self-tests pass in the fresh clone.

## Not in scope

Archiving the eval-workspace repo (after this merges), the docs-monorepo submodule (M3, planned), and the pin-sync/canary companions.


barryroodt:
Parking this alongside #119 while we redo the local-dev story TypeScript-first (cross-platform matters once this repo is open-sourced, and the bash + macOS-specific glue doesn't get us there). The core piece here (SUPABASE_MCP_SERVER_PATH + tests) is exactly what the new PR re-lands, minus the submodule, and the review feedback from this thread carries over. Branch stays for reference.

PR #119, opened 2026-07-24
M3: docs monorepo as a sparse+partial submodule

barryroodt:
Stacked on #118 (M2 fold) — retarget to `main` after it merges.

Implements M3 from the migration plan agreed in the workspace-layout thread: the docs monorepo becomes `submodules/supabase`, a **sparse+partial submodule** pinned by evals, replacing the managed `clone-docs` flat clone. All three source repos (agent-skills, mcp, supabase) now share one mechanism.

## What this buys

1. **One mechanism for all three loops** — the managed-clone plumbing (manifest `remote`, flat clone) is gone; supabase is a submodule entry like mcp.
2. **In-repo docs provenance** — the pinned SHA records exactly which docs corpus evals ran against; CI phase 6 (per-arm ephemeral index) can diff against the pin.
3. **Opt-in stays opt-in** — a plain clone leaves the gitlink empty; the seed is `mise run clone-docs`.

## Mechanics (all verified live)

- **Seed**: blob-filtered clone without checkout → cone sparse-checkout (apps/docs + 13 packages) → explicit checkout of the gitlink pin. **~56s, 311M `.git` + 622M worktree** (vs 2.9G flat clone). Re-runs refresh the sparse set only, never move HEAD.
- **`update = none`** in .gitmodules: a recursive `submodule update --init` would materialize the *entire* monorepo tree (sparse config can't be injected before the atomic clone+checkout). Verified: `git submodule update --init` prints `Skipping submodule 'submodules/supabase'`, so CI's `submodules: recursive` checkout stays safe.
- **`ignore = all`**: marker commits and pin-bump rebases in the working tree never show as dirt in evals.
- **Pin bumps**: `mise run update` keeps the fetch + rebase-markers flow and prints the deliberate record step (`git add submodules/supabase && git commit`). The generic managed-clone update flow is deleted — nothing used it.
- **Pin**: `846706b` — the base the current local index was embedded from, so the existing content DB carries over with zero re-embed drift.
- **pnpm toolchain split**: the docs monorepo pins `packageManager: pnpm@11.13.1`; evals' mise pins 10.24. Docs-loop pnpm calls run `corepack pnpm` with cwd **inside** the submodule, so the monorepo's own pin wins. (This would have bitten the flat clone on its next reinstall too.)
- **Sentry stub** (second commit): pre-existing drift — the GraphQL route calls `Sentry.flush`/`captureException`, which don't exist in the package's ESM build under standalone tsx; every request crashed. A `node:module` resolve hook no-ops the specifier. The old workspace's docs-api only worked on stale node_modules and fails today as well.

## Verification

- Seed + `apply-patches` (5 markers) + pre-push guards in the submodule; trap check green.
- Self-tests: status 19/0, hooks 16/0 (supabase patch paths now exercised), ab 23/0.
- `docs-api` serves `searchDocs` from the seeded index (1769 pages) through the submodule route; A/B dry-run resolves the docs loop with both env overrides.
- `status` renders supabase as a third nested submodule row + git version floor check (seed needs git ≥ 2.26).

## Open follow-ups (per plan)

- Docs pin-bump cadence/owner (manual `mise run update` vs renovate-style) — same ownership question as the mcp pin-sync companion.
- CI phase 6 containerizes docs-api per arm (PLAN §3e); local stays a host process — discussed, the toolchain split makes a local container net-negative.


barryroodt:
Cold run is green! Fresh clone, no containers, no volumes, zero interventions on the final pass.

| Phase | Time |
|---|---|
| Setup (clone → seeded index → serving API) | 5:59 + $0.12 one-time embed |
| Demo (`mise run ab-demo`, 2 Sonnet runs) | 2:42 |
| Teardown (back to nothing) | 0:30 |

```
baseline   passed=false checks=0/1
treatment  passed=true  checks=1/1
-> edit IMPROVED the eval (FAIL->PASS)
```

Cold here means fresh workspace + fresh Docker data; host prerequisites (mise, Docker images, keys, SSH) were already in place. Cleanup verified clean afterwards: guide reverted, canary de-embedded, corpus back to exactly 1742 pages. I think the real test is a cold run on someone else's machine.

barryroodt:
Parking this for now. The workspace glue grew very macOS-shaped (keychain, BSD sed, bash everywhere) and heavier than the job needs, and with open-sourcing on the horizon that's the wrong foundation. We're redoing the same concepts TypeScript-first in a new PR: treatment-only runs compared against the published results on main, provenance receipts, and explicit input overrides instead of submodules + patches. Branch stays for reference. A few fixes found while building this should land upstream in supabase/supabase regardless (guide checksums, the skip->purge interaction in generate-embeddings, fail-closed partner fetch).

PR #123, opened 2026-07-24
fix: retry supabase start on transient registry errors

Rodriguespn:
Currently, eval CI jobs fail whenever `supabase start` hits a Docker Hub blip: during [this run](https://github.com/supabase/evals/actions/runs/30100969319), [this job](https://github.com/supabase/evals/actions/runs/30100969319/job/89507112649) failed pulling the stack's service images (`toomanyrequests: Rate exceeded`)

#103 made the sandbox image build retry these errors, but `supabase start`, which pulls far more images, still failed on the first one.

This PR extends the same retry-with-backoff system (5s/30s/60s) to `supabase start`, so a transient registry error costs a delay instead of the run.

### DIFF
--- a/packages/sandbox/src/supabase.ts
+++ b/packages/sandbox/src/supabase.ts
@@ -29,15 +29,49 @@ export const SUPABASE_CLI_VERSION = '2.67.1';
 const SANDBOX_IMAGE_REPOSITORY = 'supabase-evals-sandbox';
 
 /**
- * Backoff schedule for retrying the sandbox image build. The build pulls
- * `node:22-slim` from Docker Hub, which intermittently answers 5xx/429 (and
- * once took out ~130 CI matrix jobs in a single blip); a short outage should
- * cost a delay, not the run. Every failure is retried — the transient
- * registry-error strings come from BuildKit/containerd internals with no
- * stable taxonomy to match against, and a deterministic failure (broken
- * Dockerfile, bad build-arg) merely wastes the ~95s schedule before failing.
+ * Backoff schedule for the registry-touching steps: the sandbox image build
+ * (pulls `node:22-slim`) and `supabase start` (pulls the stack's service
+ * images). Docker Hub intermittently answers 5xx/429 — a 502 blip once took
+ * out ~130 CI matrix jobs, and a wide `run-evals` fan-out can exhaust the
+ * runners' shared anonymous pull quota (`toomanyrequests`) all by itself. A
+ * short outage should cost a delay, not the run. Every failure is retried —
+ * the transient registry-error strings come from BuildKit/containerd/daemon
+ * internals with no stable taxonomy to match against, and a deterministic
+ * failure (broken Dockerfile, bad config) merely wastes the ~95s schedule
+ * before failing with the original error.
  */
-const IMAGE_BUILD_RETRY_DELAYS_MS = [5_000, 30_000, 60_000];
+const REGISTRY_RETRY_DELAYS_MS = [5_000, 30_000, 60_000];
+
+/**
+ * Run a registry-touching step to completion, retrying failed attempts on
+ * {@link REGISTRY_RETRY_DELAYS_MS} and returning the first ok result. On the
+ * final failure the call site's `buildError` is thrown, preserving each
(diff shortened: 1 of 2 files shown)


packages/sandbox/src/supabase.ts
```diff
@@ -319,16 +348,35 @@ async function linkSandboxToHostedPlatform(
...
   sandbox: DockerSandbox,
   includeServices?: readonly string[]
 ): Promise<void> {
-  await runOrThrow(
-    sandbox,
-    buildSupabaseStartCommand(includeServices),
-    'supabase start'
-  );
+  const command = buildSupabaseStartCommand(includeServices);
+  await withRegistryRetries('supabase start', () => sandbox.runShell(command), {
```
mattrossman: My only concern here is that this retries more than just registry-related errors. E.g. if a service gets stuck in a bad state while trying to boot up, `supabase start` won't time out for [up to 10 minutes](https://github.com/supabase/evals/blob/97d7cbde76f65934476b2a518cd002883a7631dc/packages/sandbox/src/docker-sandbox.ts#L25-L26), and if that keeps happening on each retry w/ backoff it could be ~47 min of wasted CI time.

Maybe could prevent this by either:
- skipping the retry on local stack timeout's [exit code 124](https://github.com/supabase/evals/blob/97d7cbde76f65934476b2a518cd002883a7631dc/packages/sandbox/src/docker-sandbox.ts#L373-L381)
- enforcing a stricter `timeoutMs` on the `.runShell()` call so a timeout at least surfaces sooner

and perhaps changing the withRegistryRetries name to clarify that it's not only retrying registry-related errors.

packages/sandbox/src/supabase.ts
```diff
@@ -29,15 +29,49 @@ export const SUPABASE_CLI_VERSION = '2.67.1';
...
+  for (let attempt = 0; ; attempt++) {
+    const result = await run();
+    if (result.ok) return result;
+
+    const delayMs = REGISTRY_RETRY_DELAYS_MS[attempt];
+    if (delayMs === undefined) throw options.buildError(result);
+    console.warn(
+      `[sandbox] ${label} failed ` +
+        `(attempt ${attempt + 1}/${REGISTRY_RETRY_DELAYS_MS.length + 1}), ` +
+        `retrying in ${delayMs / 1000}s`
```
mattrossman, line 79: nit, it could take longer than this for the next retry depending on how long the beforeRetry (e.g. `supabase stop --no-backup`) takes to complete

mattrossman (approved):
Would just give [this one](https://github.com/supabase/evals/pull/123#discussion_r3662054697) a look

packages/sandbox/src/supabase.ts
Rodriguespn: Good point. I'm happy with skipping the retry on local stack timeout. If we get that, that's most likely something Supabase related that requires special attention anyway (health checks are not working, for example). I'll skip the retry on local stack timeout and change the `withRegistryRetries` name to be more clear about this function's job. Thanks for the recommendation!

packages/sandbox/src/supabase.ts
Rodriguespn: Let me try to make one dependent on the other

packages/sandbox/src/supabase.ts
Rodriguespn: Done in a11c437. `supabase start` now skips the retry when the local stack times out (coreutils `timeout` exit 124/137) via a new `isTerminal` predicate on the retry helper — a wedged service is surfaced immediately instead of burning the full boot timeout on each of the 4 attempts. Also renamed `withRegistryRetries` → `withTransientRetries` and documented that it retries every non-ok result, not just registry errors. Added a unit test for the no-retry-on-timeout path.

packages/sandbox/src/supabase.ts
Rodriguespn: Done in a11c437. The backoff now counts `beforeRetry` against the delay: it captures a `readyAt = now + delayMs` before running `supabase stop`, then sleeps only the remainder. So the next attempt starts ~delayMs after the failure regardless of how long the stop takes, and the "retrying in Ns" message holds.

PR #124, opened 2026-07-24
feat(eval): Kimi K3 through OpenCode via the Vercel AI Gateway

Rodriguespn:
Runs Moonshot's **Kimi K3 through OpenCode, served by the Vercel AI Gateway**. Stacks on top of two PRs:

- **OpenCode harness** — #56
- **Vercel AI Gateway** — #120

## Stacking / review order

This is the third PR in the chain: **OpenCode (#56) → AI Gateway (#120) → this → Vercel Sandboxes (#114)**.

Because it needs *both* OpenCode and the Gateway, and GitHub PRs have a single base, it currently targets **`explore/ai-gateway-vendor-provider`** (the gateway branch) and **also carries the OpenCode harness commits**. Once #56 and #120 both land on `main`, I'll rebase this onto `main` so the diff collapses to just the Kimi delta below.

## The Kimi delta (what's new beyond #56 + #120)

- **Moonshot (Kimi) provider** for OpenCode — `moonshotai/*` ids resolve to `MOONSHOT_API_KEY`.
- **OpenCode → AI Gateway routing** — `opencodeAgent({ gateway: true })` writes a custom OpenAI-compatible provider into `OPENCODE_CONFIG` pointed at the gateway `/v1` surface and addresses the model as `vercel-ai-gateway/<vendor>/<model>` (OpenCode splits provider/model on the first `/`, keeping the gateway slug intact). The key rides in the provider config; no vendor key env is set.
- **`opencode-kimi-k3` experiment** (`suite: ['benchmark']`) running `moonshotai/kimi-k3` via the gateway.
- **Config typed against `@opencode-ai/sdk`** (type-only devDep, pinned to the installed CLI version) so config-shape drift fails to compile.

## Validation

✅ Verified **end-to-end in CI through the gateway**: [run 30104060065](https://github.com/supabase/evals/actions/runs/30104060065) — `opencode-kimi-k3 × investigate-realtime-001-subscribed-no-events` **PASS 6/6**, with `ANTHROPIC_API_KEY` withheld and only `AI_GATEWAY_API_KEY` present, so a passing eval proves the gateway carried the traffic. Unit tests cover config + exec routing (gateway vs direct).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #120, opened 2026-07-24
feat: Vercel AI Gateway as an opt-in alternative to per-vendor API keys

Rodriguespn:
Adds the [Vercel AI Gateway](https://vercel.com/ai-gateway) as an opt-in routing path for all the current three agent harnesses (`claude-code`, `codex`, `ai-sdk`), living side by side with the existing per-vendor keys. 

One env var (`AI_GATEWAY_API_KEY`) unlocks every vendor, including ones we hold no direct key for (the included Gemini experiment below proves it).

## How to switch between the two paths

| Level | Direct (default) | Gateway |
|---|---|---|
| **Whole run (CI)** | default | dispatch **Refresh eval results** with `run_through_gateway: true`, or add the **`run-evals-through-gateway`** label to a PR (⚠️ CAREFUL: triggers the full benchmark run, gateway-routed) |
| **Whole run (local)** | default | `RUN_THROUGH_GATEWAY=true pnpm eval -- …` |
| **Single experiment** | omit the option / bare model ids | `gateway: true` + slug (CLI agents) or `gateway("vendor/model")` (ai-sdk); `gateway: false` pins direct even under the flag |

## Validation

`scenario`: `build-cli-002-declarative-schema`

| Matrix job (eval `build-cli-002-declarative-schema`) | Result |
|---|---|
| `claude-code-sonnet-5` (existing direct experiment, env-flipped) | ✅ 4/4 (288s) |
| `codex-gpt-5.4-mini` (existing direct experiment, env-flipped) | ✅ 4/4 (431s) — also confirms the local 3/4 was model variance |
| `gateway-ai-sdk-gemini-3.6-flash` (suiteless, named explicitly) | ✅ 4/4 (284s) |

Check the run results and the artifacts [run 30093344664](https://github.com/supabase/evals/actions/runs/30093344664)

Ref [AI-965](https://linear.app/supabase/issue/AI-965/explore-vercel-ai-gateway-for-eval-model-requests)

PR #56, opened 2026-06-24
feat(eval): add OpenCode agent harness

Rodriguespn:
Adds `opencode` as an agent. `opencode` runs headless via `opencode run <msg> --format json --dangerously-skip-permissions` in the Docker sandbox.

## Validation 
version: opencode `v1.15.7`
scenarios:
- Tools mode: `investigate-db-001-table-row-counts`
- Local-stack mode: `build-cli-001-bootstrap-app`

| Experiment | Checks | Stop reason | Provider metadata |
|---|---|---|---|
| `opencode-claude-sonnet-5` | 3/3 ✅ | `stop` | anthropic |
| `opencode-gpt-5.4-mini` | 3/3 ✅ | `stop` | openai |
| `opencode-gemini-flash` | 3/3 ✅ | `stop` | google |

The results were the same for both scenarios

Note: Tested with the models above but we don't have to go with these three + Kimi K3

Ref [AI-964](https://linear.app/supabase/issue/AI-964/ai-spike-for-opencode-support-in-evals), [AI-967](https://linear.app/supabase/issue/AI-967/add-kimi-k3-to-public-evals-benchmark)

PR #126, opened 2026-07-24
feat: add OpenCode harness running Kimi K3 through the Vercel AI Gateway

mattrossman:
Adds `opencode` as a CLI agent harness plus two experiments running Kimi K3 through the [Vercel AI Gateway](https://vercel.com/ai-gateway): `opencode-kimi-k3` and `opencode-kimi-k3-no-skills`. Consolidates parts of #56, #120, and #124 per [this thread](https://supabase.slack.com/archives/C0BAGJBL49E/p1784919389986449?thread_ts=1784890581.560609&cid=C0BAGJBL49E). AI gateway is used only for OpenCode w/ the new experiments. Uses OpenCode's [native `vercel` provider](https://vercel.com/docs/ai-gateway/coding-agents/opencode), so the only credential is `AI_GATEWAY_API_KEY` which is already configured in repo secrets.

The CLI is pinned to 1.18.5 and must stay >= 1.17.0. Older versions don't await the run event loop ([opencode#31389](https://github.com/sst/opencode/pull/31389)) and intermittently exit mid-step with no final report, which was failing about half of sandbox runs regardless of model.

Failures like that used to be invisible: results JSON just showed an empty transcript with `stoppedReason: "error"`. Running with `--debug` now prints opencode's actual errors, which is how this was diagnosed:

```
OpenCode error event: {"name":"UnknownError","data":{"message":"Unexpected server error. Check server logs for details.","ref":"err_1ffd1f34"}}
```

Incidental fix: eval refresh jobs now fail when the experiment writes no result instead of green-skipping (surfaced by a missing `AI_GATEWAY_API_KEY` in the workflow env, now also wired up).

To test locally, add `AI_GATEWAY_API_KEY` to `.env` (create one in the [evals Vercel project](https://vercel.com/supabase/evals) under AI Gateway → API keys), then:

```sh
pnpm eval -- --experiment opencode-kimi-k3 --eval investigate-db-001-table-row-counts
```

Closes AI-967, AI-964, AI-965


.env.example
```diff
@@ -1,3 +1,10 @@
-# AI SDK Core direct-provider credentials.
+# Claude Code harness and Anthropic ai-sdk experiments.
```
mattrossman, line 1: Tweaked comments here to explain purpose of each, since we no longer primarily use these for AI SDK

packages/core/src/agents/opencode/parser.ts
```diff
@@ -0,0 +1,234 @@
...
+
+function recordToEvents(data: Record<string, unknown>): TranscriptEvent[] {
+  const type = str(data.type);
+  if (!type) return [];
+  const timestamp = toISO(data.timestamp);
+
+  if (type === 'error') {
+    const error = isRecord(data.error) ? data.error : undefined;
+    const message =
+      str(error?.message) ??
```
Rodriguespn: Checked opencode github repo and `error?.message` is always `undefined` at opencode `v1.18.5`, so this falls through to `JSON.stringify(data.error)` and surfaces a raw blob instead of the message (exactly what you pointed out in the PR ).

The `run --format json` error envelope is `{type, timestamp, sessionID, error}` where `error` is a union keyed by `name` with the text nested under `data.message`:
 
- emit: [`run.ts#L784`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/cli/cmd/run.ts#L784) + envelope [`run.ts#L678-L691`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/cli/cmd/run.ts#L678-L691)
 
- type: [`types.gen.ts#L272-L278`](https://github.com/sst/opencode/blob/v1.18.5/packages/sdk/js/src/v2/gen/types.gen.ts#L272-L278) (note [`MessageOutputLengthError#L280-L285`](https://github.com/sst/opencode/blob/v1.18.5/packages/sdk/js/src/v2/gen/types.gen.ts#L280-L285) carries no `message`)

I suggest extending this check to `str(error?.data?.message) ?? str(error?.name) ?? str(error?.message) ?? …`. 

nit: Also fix the header-comment example on line 12, which documents the wrong `{"error":{"message"}}` shape

packages/core/src/agents/opencode/runner.ts
```diff
@@ -0,0 +1,209 @@
...
+        await sandbox.exec(`mkdir -p ${SCRATCH}`);
+        await writeSandboxFile(
+          sandbox,
+          OPENCODE_CONFIG_PATH,
+          buildOpencodeConfig(mcpServers)
+        );
+        configPrefix = `OPENCODE_CONFIG=${OPENCODE_CONFIG_PATH} `;
+      }
+
+      const flags = [
```
Rodriguespn, line 136: `run --format json` only emits `reasoning` records when `--thinking` is set (the gate is [`run.ts#L761-L762`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/cli/cmd/run.ts#L761-L762)), and for non-interactive runs `thinking` defaults to `false` ([`run.ts#L275`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/cli/cmd/run.ts#L275)). 

If we want the thinking blocks in the transcript, we should be passing the `--thinking` flag.

packages/core/src/agents/opencode/runner.ts
```diff
@@ -0,0 +1,209 @@
...
+        'run',
+        message,
+        `--model ${shellQuote(`${GATEWAY_PROVIDER_ID}/${model}`)}`,
```
Rodriguespn, line 141: opencode looks `moonshotai/kimi-k3` up in its models.dev catalog and throws `ModelNotFoundError` if it's absent. It does **not** forward unknown slugs to the gateway ([`provider.ts#L1816`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/provider/provider.ts#L1816)). This is an unpinned dependency.

This is non blocking but I would consider configuring the provider routing in `provider.vercel.models["moonshotai/kimi-k3"]` block in the config ([docs example](https://vercel.com/docs/ai-gateway/coding-agents/opencode#optional-configure-provider-routing))

At minimum, a comment noting the dependency

packages/core/src/agents/opencode/runner.ts
```diff
@@ -0,0 +1,209 @@
...
+    };
+  }
+  // Typed against opencode's own `Config` schema, so a config-shape change on a
+  // CLI bump (mcp layout) fails to compile instead of silently at runtime.
+  const config: Config = {
+    $schema: 'https://opencode.ai/config.json',
+    mcp,
+  };
+  return JSON.stringify(config, null, 2);
+}
```
Rodriguespn, line 216: nit: set `OPENCODE_DISABLE_LSP_DOWNLOAD: '1'` here as forward-insurance. 

On `write`/`edit` opencode runs a detected formatter (rewriting the file) and touches an LSP it may auto-download over the network ([`write.ts#L65`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/tool/write.ts#L65), [`lsp/server.ts#L116-L143`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/lsp/server.ts#L116)). Inert in the current suite (no eval ships a formatter config/dep, and `node_modules` is stripped on copy-in), but `build-frontend-001-todos-app` with installed deps (if the agent runs `npm install`) would silently reformat the agent's output or pull an LSP mid-run.

packages/core/src/agents/opencode/runner.ts
```diff
@@ -0,0 +1,209 @@
...
+      ].join(' ');
+
+      // `< /dev/null`: opencode run blocks on stdin otherwise, even with the
+      // message passed as an argument.
+      const command = await sandbox.exec(
+        `${configPrefix}${opencode} ${flags} < /dev/null`,
+        {
+          timeoutMs: timeoutSec * 1000,
+          // The native vercel provider reads the gateway key from this env var.
+          env: { [this.apiKeyEnvVar]: apiKey },
```
Rodriguespn, line 159: nit: set `OPENCODE_DISABLE_LSP_DOWNLOAD: '1'` here as forward-insurance. 

On `write`/`edit` opencode runs a detected formatter (rewriting the file) and touches an LSP it may auto-download over the network ([`write.ts#L65`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/tool/write.ts#L65), [`lsp/server.ts#L116-L143`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/lsp/server.ts#L116)). Inert in the current suite (no eval ships a formatter config/dep, and `node_modules` is stripped on copy-in), but `build-frontend-001-todos-app` with installed deps (if the agent runs `npm install`) would silently reformat the agent's output or pull an LSP mid-run

packages/core/src/agents/opencode/runner.ts
```diff
@@ -0,0 +1,209 @@
...
+ * `$HOME` in either position. Holds the model declaration and MCP servers.
+ */
+const OPENCODE_CONFIG_PATH = '"$HOME/.eval/opencode.json"';
+
+/**
+ * Build an opencode runner bound to one gateway model slug. A single run
+ * targets one model, so the runner resolves `modelProvider` (results metadata)
+ * from the slug's vendor prefix at build time.
+ */
+export function createOpencodeRunner(
```
Rodriguespn, line 88: Every run generates a session title via a `getSmallModel()` call, and with no `small_model` set it auto-picks from the vercel catalog by family (`gemini-flash`,  `gpt-nano`, `claude-haiku` for example, [`provider.ts#L1873`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/provider/provider.ts#L1873) / [`provider.ts#L1982`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/provider/provider.ts#L1982)) — so each run silently fires e.g. `google/gemini-3-flash` through the gateway key, a cross-vendor call we aren't measuring. It's opencode's title behavior rather than the Gateway's.

Suggestion: default `small_model` to the run's own model in `createOpencodeRunner` (`vercel/<model>`, overridable via an `opencodeAgent({ smallModel })` arg) so the small model is the model under test. 

[config docs](https://opencode.ai/docs/config).

packages/core/src/agents/opencode/parser.ts
Rodriguespn, line 1: worth noting, no action needed: opencode always registers one built-in skill, `customize-opencode`, and exposes the native `skill` tool in both variants ([`skill/index.ts#L23`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/skill/index.ts#L23); our skills land in `.claude/skills` which is a native discovery root, [`skill/index.ts#L21`](https://github.com/sst/opencode/blob/v1.18.5/packages/opencode/src/skill/index.ts#L21)). If the model ever calls `skill(name="customize-opencode")`, we'll have one skill loading reference, even in the no-skills run. 

[opencode skills docs](https://opencode.ai/docs/skills).

Rodriguespn (changes requested):
I let Opus do intensive search through Opencode `v1.18.5` and docs and it flagged me some Opencode specific behaviors that i left as comments. In general the implementation is solid but there are still some caveats worth taking a second look at

packages/core/src/agents/opencode/parser.ts
mattrossman: I updated it to extract the `error?.data?.message` as you mention, but also preserve `error?.name` where possible instead of only showing it as a fallback. Tweaked the template format some too, so now it should read like:

```
[OpenCode] UnknownError: Unexpected server error. Check server logs for details.
```

packages/core/src/agents/opencode/runner.ts
mattrossman: Added the `--thinking` flag in https://github.com/supabase/evals/pull/126/commits/43c0997894d82488f585c100841fec25d969ceb4

In practice I notice scorers don't use this 'thinking' data (we only include 'message' and 'tool_call' [types in transcript parts](https://github.com/supabase/evals/blob/53c0032d6604b65ab67b89a9e647e0257c414bee/packages/core/src/index.ts#L151-L163)) but seems reasonable to emit that data for consistency w/ other harnesses in case we decide to score on thinking later.

packages/core/src/agents/opencode/runner.ts
mattrossman: I've added a [small comment](https://github.com/supabase/evals/pull/126/commits/cbb1e3bfefa23dd3013b6f9e141dce59d0b8712d) noting the live dependency. For now I'm skipping pinning the provider because we'd have to pin resolved data like `reasoning`, `tool_call` support, `cost`, etc. otherwise it uses [these fallbacks](https://github.com/anomalyco/opencode/blob/v1.18.5/packages/opencode/src/provider/provider.ts#L1450-L1491), and that data I'm not confident in maintaining in case the catalog updates for some legitimate reason. If we start to see impacts of these `ModelNotFoundError` in practice, we can revisit.

packages/core/src/agents/opencode/runner.ts
mattrossman: I'm okay to skip this for now, because:
- LSP is [disabled by default](https://opencode.ai/docs/lsp/#built-in) and no evals enable it
- If we do decide to opt-in to LSP down the road, I'd expect it to behave the same as a user would experience it instead of quietly blocking downloads

So let's revisit if we decide to enable LSP how we want to balance potential network flakiness w/ user-facing harness behavior.

packages/core/src/agents/opencode/runner.ts
mattrossman: See https://github.com/supabase/evals/pull/126#discussion_r3660897431

packages/core/src/agents/opencode/runner.ts
mattrossman: Good catch. Instead of pointing `small_model` at the run's model, I'm disabling the "title" agent entirely in https://github.com/supabase/evals/pull/126/changes/e8d087987d67773c868b2e4c0dbf474bd2ac86d7 so we're not wasting LLM calls on a user-facing title that doesn't serve a purpose in headless mode.

https://opencode.ai/docs/agents/#disable
https://opencode.ai/docs/agents/#use-title

packages/core/src/agents/opencode/parser.ts
mattrossman: Good to know! Will keep this in mind in case something weird crops up in the skill activation reports. Seems unlikely that this skill will fire for our suite. I'm okay with "no-skills" runs having skills as long as it's not supabase-maintained skills.

Rodriguespn (approved):
LGMT. Thx for pushing for this

PR #128, opened 2026-07-27
TypeScript local-dev runner: pnpm local (run/compare/experiments/docs)

barryroodt:
Replaces the parked #118/#119 approach with a TypeScript-first, cross-platform implementation of the same concepts. No bash, no macOS keychain, no submodules, no patches, no git mutation anywhere.

## What this adds

```
pnpm local run <eval...>     [--experiment <id>] [--runs N] [--mcp <path>] [--content-api <url>]
pnpm local compare <eval...> [same]   # + diff vs the latest published result on origin/main
pnpm local experiments                # agent/model/effort + published-baseline availability
pnpm local docs <up|seed|api|down> --docs <path-to-supabase-monorepo>
```

**Concept**: treat the three agent inputs as explicit overrides instead of managed clones. Skills = the tree in this repo. MCP = your own checkout, built, passed via `--mcp` (re-lands #118's core `SUPABASE_MCP_SERVER_PATH` resolution + its 12 tests, including the review fixes from that thread — thanks @Rodriguespn). Docs = a content GraphQL API served from your own supabase/supabase checkout via `pnpm local docs`, passed via `--content-api`.

**`compare` is a screen, not causal proof** — the published arm ran in the scheduled CI world. The output says so, and the receipt records the published result commit, its parent (the ref the scheduled run built on), and its age, so the confound is explicit rather than silent. `run` needs no published baseline at all (custom evals work).

**Receipts**: every run writes `results-local/<eval>.treatment.json` with host SHA + dirty state, the mcp override's git state, and the content-api URL.

**Pre-spend gates**: eval metadata parses via `parseEvalMarkdown`, experiment existence and `--mcp` path are validated — all before any model call.

**Docs loop is deliberately minimal**: full seed only, run through the docs app's own embeddings pipeline into a generated `.local-docs` workdir with its own project id and a 443xx port block (off the evals local-stack range, and below the macOS ephemeral port range — a transient outbound socket stealing a listen port is a real failure mode we hit). Incremental re-embed is deferred until the upstream pipeline fixes land in supabase/supabase (guide checksums are never set, so incremental re-embeds every guide; a source skipped for missing creds gets its still-valid rows purged).

## Verified

- `test:local` smoke: 27 checks, zero model/docker cost (fakes the run, reads real published baselines)
- core 115/115, sandbox 39 passed, typecheck clean, biome clean
- Docs loop proven live end to end: stack boots from the generated workdir, the standalone content API serves the docs route under tsx (with a Sentry no-op stub — the real package crashes outside Next instrumentation), and `searchDocs` returns ranked results against a seeded index
- `pnpm local experiments` reads experiment configs via dynamic import (accurate model/effort, no regex)


barryroodt:
Added [docs/local-workflows-design-brief.md](https://github.com/supabase/evals/blob/local-ts/docs/local-workflows-design-brief.md): a self-contained design brief for a single-page HTML presentation of the three workflows (skills / mcp / docs), for future reference.

barryroodt:
Carrying forward the pin-provenance reasoning from #109 (now closed in favour of this PR), since it outlives the mechanism that PR proposed.

`MCP_SERVER_VERSION` is a bare version string, and npm publishes no `gitHead`, so nothing in the repo ties the fixture surface to an mcp source revision. The release-please tag (`mcp-server-supabase-v0.8.1`) is the closest thing we have to that link, which is what #109 was pinning a submodule at.

It isn't academic. platform-lite tracks the pinned version, so a local build from a newer line can call endpoints the fixture doesn't serve: `get_logs` moved from `logs.all` to `logs` in 0.9.0, which 404s against a fixture sitting at 0.8.x. That's why `--mcp` prints a drift warning when the build's package version differs from the pin, and why the advice in AGENTS.md is to judge those runs by tool-call activation rather than pass or fail alone.

Cheapest follow-up, if we want it: record the release tag next to the version, so the fixture surface has a source revision to point at without carrying a submodule.

README.md
```diff
@@ -78,6 +78,65 @@ Start the web app development server:
...
+  pnpm local docs seed        # full embed via the docs app's pipeline (~$0.12 OpenAI; asks first)
+  pnpm local docs api         # keep running in a separate terminal
+  # --content-api needs a local mcp build too: SUPABASE_CONTENT_API_URL support
+  # is merged (supabase/mcp#343) but unreleased, so the published server ignores
+  # it and search_docs would silently hit production docs. Refused pre-spend.
+  pnpm local run <eval-id> --content-api http://127.0.0.1:3001/docs/api/graphql --mcp <mcp-checkout>
+  ```
+
+  **Known limitation — `docs seed` needs a docs checkout containing
+  [supabase/supabase#48364](https://github.com/supabase/supabase/pull/48364).**
```
raulb, line 112: I personally would wait until https://github.com/supabase/supabase/pull/48364 is shipped and then we could remove this limitation.

README.md
```diff
@@ -78,6 +78,65 @@ Start the web app development server:
...
+## Local development loop (`pnpm local`)
+
+Testing a change to an agent input — a skill, a local build of
+[`mcp-server-supabase`](https://github.com/supabase/mcp), or an edited docs
+page — against the evals, without touching git state:
+
+```bash
+pnpm local run <eval-id> [--experiment <id>] [--mcp <path>] [--content-api <url>]
+pnpm local compare <eval-id> [same flags]   # + diff vs the latest published result on main
+pnpm local experiments                      # list experiments + published-baseline availability
```
raulb, line 90: Would it make sense to do something like `pnpm run ... / pnpm compare ...` based on the configured environment? i.e.: local by default via some env var or flag (`--env local`)?

raulb: I realize this is a big refactor so I'll stop here

README.md
barryroodt: agreed 100% - once that PR has been merged we can remove this line completely.

README.md
barryroodt: I like the idea, let me think on that. As you say, might be a big refactor, but let me look at it before concluding that :)

barryroodt:
Split into a reviewable stack, so this PR is now the umbrella. Nothing here is abandoned, the same code lands across five PRs in dependency order:

| PR | What | Size |
|---|---|---|
| #193 | `core`: launch a local mcp build via `SUPABASE_MCP_SERVER_PATH` | 7 files, +338 |
| #194 | `pnpm local run` + provenance receipts + pre-spend gates | 7 files, +802 |
| #195 | `pnpm local compare` against the published baseline | 2 files, +240 |
| #196 | `pnpm local docs` + the `--content-api` gate | 11 files, +721 |
| #197 | docs prose | 3 files, +174 |

Each targets the one before it, so review top-down and merge in order. #193 and #197 are quick passes, which leaves three PRs of real substance.

Why this shape: the three concerns want different reviewer mindsets (container/mount semantics, CLI ergonomics, subprocess orchestration), and they turned out to be genuinely separable. The runner is coupled to the core override only through two env var *names*, and `local-docs.ts` does not import the runner at all.

#193 is the piece worth landing first regardless of the rest, since it re-lands #118 and is useful on its own with no CLI.

Two changes from what was here:

- **Three review findings fixed** since the last round: the `%H %P %cI` split that produced `NaN` on merge commits, the `--content-api` gate probing the env var rather than the flag it actually passes, and the "merged, unreleased" wording that stopped being true when mcp v0.10.0 shipped. Each has tests.
- **The single-page HTML design brief is dropped**, not carried forward. It was never referenced from anything in the repo and now lives outside it.

The stack reproduces this branch exactly, minus that brief. I verified that as an acceptance check rather than assuming: diffing the tip of #197 against `local-ts` shows the brief deletion and nothing else.

Closing this once #197 lands. @raulb's two threads here are still the open questions, in particular the `--env local` idea, which I think is a follow-up rather than something to fold in now.

PR #129, opened 2026-07-27
feat: supabase-js SDK evals (auth flow, relational report, key migration)

mandarini:
Adds four evals for supabase-js: dedicated coverage for the SDK's two most-used surfaces (auth and the PostgREST query builder), a regression canary for the API key migration, and a second Data API scenario probing an SDK-adoption gap the first one turned up.

- **`build-auth-001-email-password-flow`** (benchmark): signup/sign-in/profile flow against stubbed app code. Gates on agents reaching for `@supabase/supabase-js` unprompted (the `build-functions-005` pattern); the profile display name only survives if sent as signup user metadata.
- **`build-dataapi-001-relational-report`** (benchmark): backend reporting script over a seeded relational schema (embedded relations, aggregation). Gates on using the SDK query builder via the Data API rather than psql or a raw Postgres driver.
- **`build-dataapi-002-restock-alert-report`** (benchmark): same shape as `build-dataapi-001` — bare prompt, empty `package.json`, no SDK named — but a different schema/domain (inventory restock alerts vs. a sales report). Added after refreshed results on `build-dataapi-001` showed every claude-code variant (4/4) reaching for `@supabase/supabase-js` unprompted while every codex variant (0/4) hand-rolled raw HTTP instead, despite otherwise-correct, RLS-safe implementations. One scenario isn't enough to tell a real model tendency from an artifact of that specific prompt, so this companion checks whether the split generalizes.
- **`resolve-sdk-001-legacy-key-migration`** (regression): seeds a working app on the legacy demo JWTs; agents must migrate it to `sb_publishable_`/`sb_secret_` keys without breaking behavior or leaking the secret key into client code.

All four were sanity-checked locally against a real stack on the pinned CLI (2.109.1): migrations apply, reference solutions pass every scorer check, and the checks discriminate (missing signup metadata, publishable-key impersonation, leftover legacy keys, and SDK vs. raw-HTTP implementations all fail as intended). A scoped `.gitignore` exception ships the key-migration eval's seeded `.env`, which holds only the well-known local demo keys.

**Fix along the way:** `build-dataapi-001-relational-report` and `resolve-sdk-001-legacy-key-migration` initially failed 100% of agents identically — a harness bug, not an agent gap. Both had omitted `gotrue` from their `services:` frontmatter, so the local stack's auth service never started and `supabase status` couldn't emit `PUBLISHABLE_KEY`/`SECRET_KEY`. Fixed by adding `gotrue` to `services:` (matching `resolve-dataapi-002-secure-default-grants`'s pattern); results refreshed and now produce real signal.

`build-dataapi-002-restock-alert-report` still needs a live run (no API keys in this pass) to confirm whether the adoption split holds.


evals/build-auth-001-email-password-flow/EVAL.ts
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * GATING: some app code file must genuinely import @supabase/supabase-js —
+ * we match the quoted module specifier, not a bare mention in a comment.
+ */
+async function sdkUsageCheck(ctx: LocalStackEvalContext): Promise<CheckResult> {
+  const NAME = 'implementation uses @supabase/supabase-js';
+  const scan = await ctx.exec(
+    `grep -rlE --exclude-dir=node_modules --include='*.mjs' --include='*.js' --include='*.cjs' --include='*.ts' ` +
+      `"['\\"](npm:)?@supabase/supabase-js" ${APP_DIR} || true`
```
barryroodt: `"['\"](npm:)?@supabase/supabase-js"` requires an opening quote but nothing after the package name, so it also matches `@supabase/supabase-js-not-real`. The comment above says we match the quoted specifier rather than a bare mention, and `grep -rl` over all of `app/` means a commented-out line or a dead helper satisfies it. Since this check is what gates every failure here, could we resolve the import from the scored entry instead?

Same pattern at [`build-dataapi-001/EVAL.ts:174`](https://github.com/supabase/evals/blob/a3024b395bb58cc52d4da494d2d9b29a9a80373f/evals/build-dataapi-001-relational-report/EVAL.ts#L174) and [`build-dataapi-002/EVAL.ts:162`](https://github.com/supabase/evals/blob/a3024b395bb58cc52d4da494d2d9b29a9a80373f/evals/build-dataapi-002-restock-alert-report/EVAL.ts#L162).

evals/resolve-sdk-001-legacy-key-migration/EVAL.ts
```diff
@@ -0,0 +1,230 @@
...
+      `select title from public.posts where published order by title asc`
+    );
+    const expectedTitles = publishedRows.map((row) => row.title);
+    const { rows: draftRows } = await ctx.query(
+      `select count(*)::int as n from public.posts where not published`
+    );
+    const expectedDrafts = Number(draftRows[0]?.n ?? -1);
+
+    // 1. The public script still lists published posts (client-key path).
+    const posts = await ctx.exec(`cd ${APP_DIR} && npm run -s posts`, {
```
barryroodt, line 62: The four checks can all pass without any key migration happening. `publishableKey` and `secretKey` get read at lines 27-28, then only used for the guard, never injected into `npm run posts` / `npm run stats` and never compared against what the app ends up using.

So a script that prints the seeded titles and draft count, with no Supabase call at all, passes: check 1 and 2 match on output, check 3 sees no legacy marker, check 4 sees no secret reference. Worth asserting the new keys are actually in use, maybe by injecting them and failing when the app ignores them.

evals/build-auth-001-email-password-flow/EVAL.ts
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    const written = await writeDriver(ctx);
+    if (!written.ok) {
+      return fail(
+        'installed the eval driver',
+        written.stderr.trim() || written.stdout.trim()
+      );
+    }
+    const run = await ctx.exec(
```
barryroodt, line 91: This runs the driver without installing app deps first, while the other three scorers do a guarded install before their run. A solution that adds `@supabase/supabase-js` to `package.json` but leaves no `node_modules` behind dies here on `ERR_MODULE_NOT_FOUND`, so the score depends on whether the agent happened to install rather than on what it submitted. Same guarded install before this line?

evals/build-dataapi-001-relational-report/EVAL.ts
```diff
@@ -0,0 +1,214 @@
...
+      return fail(
+        'read stack config from `supabase status`',
+        `missing API_URL/SECRET_KEY — is the stack running on a new-enough CLI? got keys: ${Object.keys(status).join(', ')}`
+      );
+    }
+
+    // Be generous about a missing install step; the eval is about the report,
+    // not npm. A no-op when the agent already installed dependencies.
+    await ctx.exec(
+      `cd ${APP_DIR} && [ -d node_modules ] || npm install --no-audit --no-fund --silent || true`,
```
barryroodt: The `|| true` swallows install failures, and the result isn't inspected. A registry hiccup or a bad lockfile then surfaces later as `ERR_MODULE_NOT_FOUND` from the script, which reads as an agent mistake in the results. Could we keep the install and push an explicit failed setup check with its stderr when it breaks?

Same at [`build-dataapi-002/EVAL.ts:58`](https://github.com/supabase/evals/blob/a3024b395bb58cc52d4da494d2d9b29a9a80373f/evals/build-dataapi-002-restock-alert-report/EVAL.ts#L58) and [`resolve-sdk-001/EVAL.ts:39`](https://github.com/supabase/evals/blob/a3024b395bb58cc52d4da494d2d9b29a9a80373f/evals/resolve-sdk-001-legacy-key-migration/EVAL.ts#L39).

evals/build-dataapi-002-restock-alert-report/PROMPT.md
```diff
@@ -0,0 +1,32 @@
...
+  - sdk
+services:
+  - gotrue
+  - kong
+  - postgrest
+projectRunning: true
+motivation: >-
+  build-dataapi-001-relational-report found every codex variant (0/4) skips
+  @supabase/supabase-js for a bare backend Data API script, hand-rolling raw
+  HTTP instead, while every claude-code variant (4/4) reached for it
```
barryroodt, line 19: The 4/4 claim for claude-code is already stale at this head, `build-dataapi-001` shows `claude-code-sonnet-5-no-skills` failing, so it's 3/4. Since the split is the stated reason for adding this scenario, the number matters.

Also worth a user-facing citation here alongside the internal comparison, `CONTRIBUTING.md` asks for evidence from the user journey and right now this cites only the sibling eval. Same number in [`README.md:5`](https://github.com/supabase/evals/blob/a3024b395bb58cc52d4da494d2d9b29a9a80373f/evals/build-dataapi-002-restock-alert-report/README.md#L5).

.gitignore
```diff
@@ -2,6 +2,8 @@ node_modules/
 dist/
 .env
 .env*.local
+# eval seed data may include a .env with well-known local demo keys
+!evals/*/local/**/.env
```
barryroodt: `!evals/*/local/**/.env` un-ignores `.env` for every eval, including ones that don't exist yet, so a future seed can commit a real credential without tripping the guard. `deploy-functions-001` shows a forced add works for an exact fixture. Scope it to `evals/resolve-sdk-001-legacy-key-migration/local/app/.env`?

evals/build-dataapi-001-relational-report/local/supabase/.temp/cli-latest
```diff
@@ -0,0 +1 @@
+v2.110.0
```
barryroodt: These look like generated CLI state that got picked up by the add. No other eval tracks `.temp/` or `.branches/`, and this one records `v2.110.0` while the prompt pins `2.109.1`. Since `local/` is copied into the agent workspace they'd land as seed data, so probably worth deleting all four (`.temp/cli-latest` and `.branches/_current_branch` here and under `build-dataapi-002`) and ignoring both dirs.

evals/build-dataapi-001-relational-report/EVAL.ts
```diff
@@ -0,0 +1,214 @@
...
+      { timeoutMs: 180_000 }
+    );
+
+    const run = await ctx.exec(
+      `cd ${APP_DIR} && SUPABASE_URL="${apiUrl}" SUPABASE_SECRET_KEY="${secretKey}" node ${REPORT}`,
+      { timeoutMs: 60_000 }
+    );
+    const actual = parseReport(run.stdout);
```
barryroodt, line 86: Minor: this parses stdout without looking at `run.ok` or the exit code, so a script that prints the right JSON and then exits 1 still passes. Same at [`build-dataapi-002/EVAL.ts:66`](https://github.com/supabase/evals/blob/a3024b395bb58cc52d4da494d2d9b29a9a80373f/evals/build-dataapi-002-restock-alert-report/EVAL.ts#L66).

barryroodt (commented):
These four are in good shape, and pulling the expected numbers out of the seeded migrations instead of hardcoding them is the right call. The auth driver holding one session across signup and both signin attempts is nice too.

One thing to sort before merge: `implementation uses @supabase/supabase-js` is the only failed check in all 13 benchmark failures, and the grep behind it doesn't prove SDK use. The pattern requires a quote before the package name but nothing after it ([`EVAL.ts:198`](https://github.com/supabase/evals/blob/a3024b395bb58cc52d4da494d2d9b29a9a80373f/evals/build-auth-001-email-password-flow/EVAL.ts#L198)), so `@supabase/supabase-js-not-real` matches, and an unused import passes just as well as a real one. So those three evals currently score whether a string appears somewhere under `app/`, which makes the codex/claude split hard to read.

My call is to fix the detection rather than name the SDK in the prompt, since unprompted reach is the thing you're measuring and that only means something if the gate is real. Resolving the import from the scored entry, or instrumenting module load, would both do it.

Rest is inline. The `resolve-sdk-001` one is worth a look, a hardcoded stub currently passes it 4/4.

mandarini:
@barryroodt thanks for the review! addressed your comments. Can you please take a look?

evals/resolve-sdk-001-legacy-key-migration/EVAL.ts
barryroodt: That closes it, the hardcoded stub fails both new checks now. One fix: `stillLooksValid` passes as long as the output still parses, so a valid migration that handles the API error and prints `[]` reads as hardcoded. Compare against the expected titles and count instead.

Two nits in the same function: neither `writeFile` call checks its result, and the corrupt target is only `.env` or the entry file, so a key hardcoded in a helper module lands on `could not find the live key on disk`.

evals/build-auth-001-email-password-flow/EVAL.ts
barryroodt: Prefix match is gone, thanks. Scope is still all of `app/`, so an unused import in the entry counts as SDK use. That moves none of the current rows, so I'm not blocking on it.

evals/build-auth-001-email-password-flow/EVAL.ts
barryroodt: Fixed. Nit: `[ -d node_modules ]` skips the install whenever the dir exists, so an agent that installed before adding the SDK still dies on `ERR_MODULE_NOT_FOUND`.

barryroodt (commented):
Nice, this is a lot tighter. Two things before merge.

The new key-dependency checks on `resolve-sdk-001` only ask whether the corrupted run still prints parseable JSON, so a valid migration that handles the API error and prints `[]` fails with a misleading "hardcoded" result. Comparing against the expected titles and draft count you already compute fixes it.

`build-dataapi-002-restock-alert-report` has no result rows at this head, down from 8 (2 passing) at a3024b39, so I think the rebase dropped them. The rows that remain predate 488a985, so one refresh after that fix covers both.

Everything else from the first review is addressed. Nits are inline.

evals/resolve-sdk-001-legacy-key-migration/EVAL.ts
barryroodt: The refreshed rows split into two different problems.

`stats` is the false failure I described: `still produced {"drafts":0}` against an expected 2, so comparing to the expected count passes it.

`posts` is something else. It returned the correct three titles with a corrupted publishable key, so comparing to expected won't rescue that one. Two candidates I can't separate from the results alone: the key sits in both `.env` and the script and only the first hit gets corrupted (the check picks `.env` and stops there), or the local stack doesn't reject an unknown publishable key on anon reads, which would mean that half of the check can't work locally. Do you still have the workspace from that run?

barryroodt:
Your refresh landed while I was writing that, so scratch my second point. `build-dataapi-002` has its 10 rows back, and `resolve-sdk-001` now shows 6 checks. dataapi-002 came out 4/4 claude-code, 0/4 codex, 0/2 opencode, so the split does generalize.

It confirms the first point though, both resolve-sdk rows now fail on the two new checks. More on that thread.

mandarini:
@barryroodt pushed one more commit!

barryroodt (approved):
Nice, that's all of it. Corrupting the URL instead of the key is a better call than what I suggested.

`resolve-sdk-001` is 2/2 on the refreshed rows, both new checks reporting `broke as expected against an unreachable project URL`.

PR #130, opened 2026-07-27
feat: count shell fetches as docs activation, read Codex's `web_search` action

mattrossman:
Changes how we track docs activation for Codex:
- stdout-facing shell commands like `curl` and `wget -O -` are now counted as docs calls (with content)
- The [`web_search`](https://developers.openai.com/api/docs/guides/tools-web-search) tool's `action` payload is now used to determine whether the model opened a page, instead of inferring this based on a URL-shaped `query`
- Docs calls where `hasContent` is undefined are rendered with a distinct question mark instead of falling back to looking like content-ful calls

Example of the new `SHELL` docs activation tracking in the "cli 001: bootstrap app" eval run from the preview app:

CleanShot 2026-07-28 at 17 40 55@2x

Another example where the `hasContent` couldn't be determined:

CleanShot 2026-07-28 at 17 40 47@2x

## Motivation

Our skill tells agents to fetch `https://supabase.com/changelog.md` and in several of our runs, Codex achieves this with `curl` which we're not currently counting. For comparison, Claude Code uses `WebFetch` which is counted.

This change aims to more accurately count changelog reads, among other docs pages.

The `web_search` change is less important, but still a technical accuracy thing for how we count page reads in Codex.

## What's counted

This only counts commands that write page content to stdout, since we can generally trust that means the content was added to context. It counts `supabase.com` domains only, excluding service subdomains. `resultChars` is measured after whatever the agent piped the response through, so it counts what reached the model rather than the size of the whole page. If the command has ambiguous output, we treat `hasContent` as undefined.

It doesn't count commands that don't direct to stdout (e.g. downloading a page and then reading it from filesystem). The reasoning is that it's harder to reliably track this, as there's tons of ways that our heuristics might be inaccurate. We also don't see evidence of this happening in practice to justify the effort. We can revisit that sort of tracking once we notice it impacting results.

See tests cases for more context on what's counted vs excluded.

## Impact on transcript and UI

This adds a `shell_fetch` type to `DocsCall.source` similar to existing MCP `seach_docs` and harness `web_fetch` and `web_search` sources.

In this UI this appears as a `SHELL` chip tagged on the docs call. I label it as such since it represents multiple possible shell fetch strategies, but if desired we could split into the specific binary like `CURL`, `WGET`, etc.

Shell invocations that mix multiple commands w/o piping are counted as docs calls but with `hasContent` undefined, since we can't be sure what content was printed to stdout as context. Those now render distinctly in the UI from known content-ful or content-less calls.

## On `web_search.action` change

Note that although this PR starts reading Codex's `web_search.action` payload, there's a bug in Codex preventing this helping to its full potential. Similar to [this fixed issue](https://github.com/openai/codex/pull/10408), there's inconsistency in the `action` casing between [validation](https://github.com/openai/codex/blob/rust-v0.138.0/codex-rs/exec/src/event_processor_with_jsonl_output.rs#L305-L310) ([camelCase](https://github.com/openai/codex/blob/rust-v0.138.0/codex-rs/app-server-protocol/src/protocol/v2/item.rs#L742-L745) vs. [snake_case](https://github.com/openai/codex/blob/rust-v0.138.0/codex-rs/protocol/src/models.rs#L1194-L1196)) that causes the `open_page` action (or any multi-word action) to fall back to `other`.

You can see this in practice by running:

```bash
codex exec --json --ephemeral --skip-git-repo-check \
  'Use web search to open https://raw.githubusercontent.com/openai/codex/main/CHANGELOG.md directly. Do not use curl or shell commands.' \
  | jq -c 'select(.item.type == "web_search") | {query: .item.query, action: .item.action}'
```

We can consider opening an issue in the Codex repo, and for now this change is a guard for if/when that validation issue gets fixed.

Nonetheless, the other `action` falls back to the existing URL-shaped query heuristic, so it's not regressing our results.

I've updated results for the Codex experiments on this PR, I didn't refresh other harnesses since I haven't observed them using shell to fetch web pages thusfar.

Closes AI-958


Rodriguespn (approved):
LGTM. We discussed offline the details of this so I'm in the loop. Nice catch and thkx for taking care of this

packages/core/src/docs-results.ts
```diff
@@ -48,6 +59,78 @@ function isSupabaseUrl(value: string): boolean {
...
+  return typeof call.body.command === 'string' ? call.body.command : undefined;
+}
+
+/** The docs urls a shell command writes to stdout, empty when page text won't reach the model. */
+function shellFetchUrls(command: string | undefined): string[] {
+  if (!command) return [];
+  const urls: string[] = [];
+  for (const segment of command.split(SHELL_SEGMENT_PATTERN)) {
+    const curlToStdout =
+      CURL_PATTERN.test(segment) &&
```
barryroodt: **Any command that mentions curl scores a page read.** `/\bcurl\b/` matches the word anywhere in the segment, not curl as the invoked program, so `echo "run curl https://supabase.com/changelog.md next"` comes back as `hasContent: true` with `resultChars: 47` sourced from unrelated output. In the UI that's indistinguishable from a real changelog fetch.

The guard test at `docs-results.test.ts:601` looks like it covers this, but it passes only because its echo doesn't happen to contain the word "curl". Swap `see` for `curl` in that fixture and it fails, which is probably the cheapest regression to add.

Simplest fix I can see is requiring the fetcher at a command position (segment start, after env assignments), though rejecting known non-fetchers like `echo`/`rg`/`grep`/`cat` would also do it.

packages/core/src/docs-results.ts
```diff
@@ -48,6 +59,78 @@ function isSupabaseUrl(value: string): boolean {
...
+      !CURL_NO_BODY_PATTERN.test(segment) &&
+      !segment.includes('>');
```
barryroodt: **A stderr redirect drops the fetch entirely.** `curl -fsSL https://supabase.com/changelog.md 2>/dev/null | head -50` records no calls, because `includes('>')` can't tell which fd is being redirected. That's a routine agent idiom, and it's a silent false negative on exactly the changelog case this PR is fixing.

Worth noting the check has to stay fd-aware rather than just dropping the `>` test: `>file`, `1>file` and `&>file` all capture stdout and should still disqualify, only `2>file` and `2>&1` leave the body on stdout.

packages/core/src/docs-results.ts
```diff
@@ -82,7 +165,8 @@ function isDocsRelatedCall(call: ToolCallRecord): boolean {
   return (
     call.endpoint.endsWith('search_docs') ||
     call.name === 'web_fetch' ||
-    call.name === 'web_search'
+    call.name === 'web_search' ||
+    shellFetchUrls(shellCommand(call)).length > 0
```
barryroodt, line 179: **Admitting shell calls here has a second effect beyond the docs metric.** `rehydrateTruncatedDocsResults` mutates `call.result` in place before scorers run, so a truncated Bash result now gets replaced with the full file contents for judges too, not just for docs counting. Might be what you want, it just doesn't look deliberate from here.

Smaller thing while you're in here: this predicate runs before the truncation-stub check below it, so every shell command in every eval gets split, regex scanned and URL parsed even though almost all of them fail that next check. Pulling `extractTruncatedResultPath` up first would skip nearly all of it.

packages/core/src/docs-results.ts
```diff
@@ -170,8 +254,50 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
...
+      //
+      // Only `search` actually arrives intact on CLI 0.138: exec re-parses the
+      // app-server action, which serializes camelCase (`openPage`), into a
+      // snake_case enum (`open_page`), so both page-reading variants land on
+      // the catch-all and reach us as `other`. `search` survives because it's
+      // one word in either casing. The `other` calls fall through to the
+      // url-shape branch below, same as before. See codex's
+      // event_processor_with_jsonl_output.rs (the from_value round trip) and
+      // app-server-protocol v2/item.rs vs protocol/models.rs for the two enums.
+      const action = webSearchAction(body);
```
barryroodt, line 278: **Codex protocol detail is living in a shared scorer.** `action` arrives raw here along with the 0.138 casing writeup and the fallback policy, so the next action type or casing change is a coordinated edit across `agents/codex/parser.ts` and this file. Every other harness quirk we have sits in its parser.

Normalizing at the parser would let this file drop `webSearchAction`, both action branches, the `isRecord` import and the comment. The one thing I'd keep intact is the `other` path: it currently reaches the url-shape fallback and only records a page when the query is actually url-shaped, so it shouldn't be classified as a page open unless the parser can prove one.

Boundary cleanup rather than a correctness fix, so I'd be happy to see it as its own PR, or not at all if you think the comment carries it well enough.

packages/core/src/docs-results.ts
```diff
@@ -170,8 +254,50 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
...
+
+      if (action?.type === 'open_page' || action?.type === 'find_in_page') {
+        if (!action.url || !isSupabaseUrl(action.url)) continue;
```
barryroodt: Small one: this uses `isSupabaseUrl`, so an `open_page` on `https://mcp.supabase.com/mcp` records as a docs read, while `curl` of that same host is dropped by `isSupabaseDocsUrl` at line 105 and `docs-results.test.ts:568` asserts the drop as intended. Same "agent read a docs page" signal, two domain rules. Worth picking one.

packages/core/src/docs-results.ts
```diff
@@ -196,6 +322,23 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
       });
       continue;
     }
+
+    if (call.name === 'shell') {
+      const command = shellCommand(call);
+      const urls = shellFetchUrls(command);
+      if (!command || urls.length === 0) continue;
+      // With no shell output at all, nothing from the fetch reached the model.
+      if (typeof result !== 'string' || result.length === 0) continue;
```
barryroodt, line 342: **A failed fetch is scored as a content-ful read whenever the failure never becomes `error`.** `curl` without `-f` exits 0 on a 404, so `404: Not Found` lands in `result` and produces `hasContent: true, resultChars: 14` with the URL attached. The record positively asserts a changelog read that didn't happen.

Two more ways in, if you want them covered in one go: a `command_execution` item with no `exit_code` leaves `success: undefined` (`agents/codex/parser.ts:145`), which `adapt.ts` doesn't divert to `error`, and that route defeats `-f` too. Claude Code's `tool_result` with no `is_error` reads as success the same way.

The non-zero-exit path is genuinely fine, so this is narrower than it first looks: it's the "non-empty string result proves content" assumption rather than the failure handling.

packages/core/src/docs-results.ts
```diff
@@ -196,6 +322,23 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
...
+      const ownsResult = shellFetchOwnsResult(command);
+      calls.push({
+        source: 'shell_fetch',
+        query: command,
```
barryroodt, line 351: Storing the whole command means anything in it gets committed to `eval-results.json` and rendered in the preview app. A `curl -H "Authorization: Bearer sb_secret_..."` would persist verbatim. The 13 entries at this SHA are clean, so this is precautionary rather than a live problem, but since the repo is public it might be worth stripping `-H`/`--header`/`-u` values, or just storing the matched URLs.

packages/core/src/eval-metadata.ts
```diff
@@ -261,14 +264,14 @@ export type DocsCallPage = z.infer<typeof docsCallPageSchema>;
...
-  // False for Claude Code's WebSearch (its results never include page text,
-  // only title/url). Unknown (omitted) for a Codex web_search used as a
-  // fetch: no result payload is ever exposed on that tool.
+  // for `content`), web_fetch, and an isolated shell_fetch. False for Claude
+  // Code's WebSearch, whose results never include page text. Unknown (omitted)
+  // when the available trace cannot prove either state, including Codex
+  // web_search and a shell fetch mixed with other commands.
   hasContent: z.boolean().optional(),
   pages: z.array(docsCallPageSchema),
   // Size of the result the call actually produced, in characters, an
```
barryroodt, line 278: This comment ends "Omitted only when there's no result at all", but `docs-results.ts:338` also omits `resultChars` for a mixed shell command, which is reached past the non-empty result check. Looks like an easy one to miss, since the `hasContent` comment just above did get updated for that same case.

barryroodt (changes requested):
For the sake of completeness, I had my agent run a thorough review on this, all feedback below, along with the inline comments, is a result of that run.

Nice direction, though the shell heuristic still miscounts a few common cases, all in the direction that moves the number this PR exists to make accurate.

- `echo "run curl https://supabase.com/changelog.md next"` records a page read with `hasContent: true`. The word `curl` anywhere in the segment is enough.
- `curl -fsSL <changelog> 2>/dev/null | head -50` records nothing at all.
- A 404 counts as a read, since curl without `-f` exits 0 and the error body lands in `result`.

Left those inline with the details. A regression test for each would be the useful part, more than the specific fix.

Separately on the results: they look re-run rather than re-scored. 71 of 190 rows differ, 53 of those changed a run-level field like `attempts` or `checks`, `search_docs` moved 146 to 142 even though this PR doesn't touch that branch, and 7 rows flipped pass/fail. Nothing wrong with re-running, just worth a line in the description so nobody reads those flips as effects of this change.

One structural thing I'd like your take on, and I think it's fine as a follow-up: the Codex action handling lives in `docs-results.ts` even though that module is meant to stay agent-agnostic.

packages/core/src/docs-results.ts
mattrossman: I'm okay with this, these are commands run by agents in test projects so I don't expect anything sensitive to appear. I'd like to see the actual command run for transparency about how we're tracking docs activation.

mattrossman:
> Separately on the results: they look re-run rather than re-scored

Yep, that's just a byproduct of refreshing results on the PR w/ nondeterminism. All good.

packages/core/src/eval-metadata.ts
mattrossman: I've updated the actual implementation in https://github.com/supabase/evals/pull/130/commits/766c1a90b72301409ddb401624d5463d2c6ac003 so it no longer omits resultChars for mixed shell commands (since we might still be curious to see how much is pulled into context for those commands, even if we're not sure the content is attributable to docs) and tweaked the comment here slightly for clarity.

mattrossman:
>the Codex action handling lives in docs-results.ts even though that module is meant to stay agent-agnostic.

Agreed it's not ideal, I'm okay leaving it for now because this doesn't affect the persisted results shape, and can be simplified by using the existing normalized `url` once the casing bug on Codex's end is resolved.

packages/core/src/docs-results.ts
mattrossman: This is a bit tricky to solve correctly since the 404 body might take several shapes, so for now I'm updating this to treat curl responses w/o the `-f` flag as `hasContent: unknown` in https://github.com/supabase/evals/pull/130/commits/313bd1e78d4b71484b551afb259fed75b795122a

Down the line it'd be nice to support a proper `error` channel on docs calls so we can surface calls to hallucinated pages.

packages/core/src/docs-results.ts
mattrossman: Consolidated the two conflicting helpers into one `isSupabaseApexUrl` used in both places. https://github.com/supabase/evals/pull/130/changes/5de73b910c56ab06b2792c1810b52fc5d4941ec6

packages/core/src/docs-results.ts
mattrossman: https://github.com/supabase/evals/pull/130#issuecomment-5123743609

packages/core/src/docs-results.ts
mattrossman: Good catch, fixed the ordering in https://github.com/supabase/evals/pull/130/commits/9e36add6b5faa253984a426473d35edc48457f99

packages/core/src/docs-results.ts
mattrossman: Corrected in https://github.com/supabase/evals/pull/130/commits/4da4d912347c04a914d234b05706cfbf331d55f8 with new regression tests

packages/core/src/docs-results.ts
mattrossman: Fixed in https://github.com/supabase/evals/pull/130/commits/e9d7220de03127602510f50c7a7487282a0756a2 with regression test

barryroodt (commented):
All five fixes hold, and the guard tests are the right ones. I ran `pnpm --filter @supabase-evals/core test` at e51c7b6 (101 pass) and replayed all 13 persisted `shell_fetch` commands through the new code.

One blocker: the fetcher-position fix drops a multi-line command entirely, since [`SHELL_SEGMENT_PATTERN`](https://github.com/supabase/evals/blob/e51c7b66a56c9026f0277aa945ce637f3ddb539e/packages/core/src/docs-results.ts#L42) splits on `;` and `|` but not newlines, so a `curl` on a later line is never at a segment start. A row in this PR's own results file shows it: the `set -e` / `printf` / `curl -fsSL .../functions/auth.md` run persisted 2 docs pages, current code records none, and that's the one place the committed results and the code disagree.

Adding `|\r?\n` to that pattern (its sibling [`SHELL_COMMAND_LIST_PATTERN`](https://github.com/supabase/evals/blob/e51c7b66a56c9026f0277aa945ce637f3ddb539e/packages/core/src/docs-results.ts#L43) already has it) brings the row back and keeps all 101 tests green.

mattrossman:
@barryroodt I've updated in https://github.com/supabase/evals/pull/130/changes/612cb7dfaa2a79d36e3737e3d344c2f88747f4ac to add tests + fix for multi-line commands. I'm treating `hasContent` as unknown in those cases b/c there's many possible edge cases where the command's output may not be directly attributable to a URL.

I haven't refreshed results on the PR and will likely discard them before merge because I don't want to shake up our recently published benchmark results, and it takes a while to re-run the full matrix, so I'm okay having that bundled into a separate matrix refresh PR.

barryroodt (approved):

PR #140, opened 2026-07-29
nuqs

SaxonF:
Use Nuqs for state management to allow shareable urls

Fixes https://linear.app/supabase/issue/AI-981/support-shareable-urls-for-eval-result-views


raulb (approved):
Amazing, thank you!

PR #156, opened 2026-08-02
fix(framework): resolve vite/vitest from apps/framework, not repo root

Rodriguespn:
Solves the lingering failed framework smoke test that was erroring because `project-runner.ts` looked for `vite`/`vitest` in the monorepo root's `node_modules` instead of `apps/framework`'s, where pnpm actually installs them.

**Note:** no CI workflow runs `pnpm check`/`typecheck` (only `biome format:check` does), so this failure never surfaced in Actions — only locally, and silently, since the smoke script's error output is suppressed unless run with `--debug`.

**Repro on main:**
```
pnpm --filter @supabase-evals/framework test:framework -- --debug
```
Fails with `Cannot find module '.../node_modules/vite/bin/vite.js'`.

### DIFF
--- a/.github/workflows/eval-refresh.yml
+++ b/.github/workflows/eval-refresh.yml
@@ -312,7 +312,9 @@ jobs:
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: raw-results-${{ matrix.experiment }}__${{ matrix.eval_id }}
-          path: results/${{ matrix.experiment }}/
+          path: |
+            results/${{ matrix.experiment }}/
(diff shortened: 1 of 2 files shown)


barryroodt (changes requested):
Good catch, and this passes the framework smoke test locally.

One issue before merge: each workspace now keeps a `node_modules` symlink under `results/`, and [`upload-artifact` follows symlinks](https://github.com/actions/upload-artifact/blob/043fb46d1a93c77aae656e7c1c64a875d1fc6a0a/src/shared/search.ts#L14-L21), so every frontend eval artifact will carry the framework dependencies. Could you add `!results/**/node_modules/**` to the upload step?

barryroodt (approved):

PR #160, opened 2026-08-03
chore: refresh codex-gpt-5.6 prometheus-metrics runs at medium reasoning

barryroodt:
Fixes the `describes each experiment the same way on every one of its runs` failure in [`eval-results.test.ts`](https://github.com/supabase/evals/blob/b8641b9/apps/web/src/lib/eval-results.test.ts#L203-L219), which fails on `main` today.

[#142](https://github.com/supabase/evals/pull/142) moved GPT-5.6 Sol to `reasoningEffort: 'medium'`, but the `deploy-database-001-prometheus-metrics` runs for `codex-gpt-5.6` and `codex-gpt-5.6-no-skills` were never re-run, so each experiment carried 18 runs at `medium` and 1 at `low`. Since [`getExperimentDisplay`](https://github.com/supabase/evals/blob/b8641b9/apps/framework/scripts/export-results.ts#L119-L120) takes the first matching run, that disagreement can flip a model's label depending on sort order.

Refreshed both pairs through `eval-refresh.yml` (runs [30806941644](https://github.com/supabase/evals/actions/runs/30806941644) + [30807365508](https://github.com/supabase/evals/actions/runs/30807365508)), one experiment at a time because the harness ANDs `--experiment` with `--experiment-suite`, so the two cannot share a matrix.

Only the two target rows changed, 190 rows in and 190 out. Scores held at 3/3 for both, so this is metadata and transcript churn rather than a scoring change. `pnpm --filter @supabase-evals/web test` goes 70 passed + 1 failed to 71 passed.

Worth noting this comes back on any future `reasoningEffort` change that skips a re-run. Making the label read from the live experiment config instead of the first run would close it off for good, though that is a bigger call about what a historical run should report.

### DIFF
--- a/apps/web/src/data/eval-results.json
+++ b/apps/web/src/data/eval-results.json
@@ -12560,7 +12560,7 @@
       "agent": "codex",
       "modelProvider": "openai",
       "modelId": "gpt-5.6-sol",
-      "reasoningEffort": "low"
+      "reasoningEffort": "medium"
     },
     "eval": "deploy-database-001-prometheus-metrics",
     "stage": "deploy",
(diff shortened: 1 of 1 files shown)


mattrossman:
\>the two cannot share a matrix<br><br>This part has been bugging me when refreshing results, would like to fix at some point

PR #162, opened 2026-08-03
feat: promote build-rls-003-org-roles-permissions to benchmark suite

mattrossman:
(Diff note: `+1,074 / -15` of changes are just generated results.)

## Why

Surfaced by Copple on [Slack](https://supabase.slack.com/archives/C051L8U2EJF/p1785754901030489), we want more RLS benchmarks in our evals.

## Fix

Promotes `build-rls-003-org-roles-permissions` to the benchmark suite and tweaks its prompt / scoring.

Originally this scenario also tested the soft-delete and audit log implementation pattern, for which I found motivations [FDBKIN-8277](https://linear.app/supabase/issue/FDBKIN-8277/provide-a-clear-pattern-or-template-for-implementing-soft-deletes-with) and [FDBKIN-9175](https://linear.app/supabase/issue/FDBKIN-9175/provide-built-in-audit-logs-for-tracking-user-changes-to-data) respectively, however I removed these from the scenario to keep the focus on RLS.

Other changes:

- Added `motivation`
- Simplified prompt
- Seeded a leftover "Enable read access for all users" policy similar to Studio's default template so agents have to fix the existing security hole
- Added a few more checks to try hardening the scenario e.g. no cross-org `member` roster visibility, no cross-org role privilege escalation

Agents already pass this across the board, which is against our typical criteria for adding benchmarks, but I figure we're okay here in an effort to flesh out RLS coverage.

CleanShot 2026-08-04 at 09 28 59@2x

## How to review

Look over the preview results:
https://evals-git-mattrossman-ai-1006-promote-build-rls-47a093-supabase.vercel.app/?group=eval&sheet=eval&item=build-rls-003-org-roles-permissions

Sanity check it looks reasonable.

Also glance over `PROMPT.md` and checks in `EVAL.ts` to see if we're fair in our scenario and testing.

Closes AI-1006


evals/build-rls-003-org-roles-permissions/EVAL.ts
```diff
@@ -1,227 +1,331 @@
...
+async function checkCannotSeeAnotherOrgsMembershipRoster(
+  ctx: ToolEvalContext
+): Promise<CheckResult> {
+  const result = await runAsUser(
+    ctx,
+    EDITOR_A,
+    `SELECT user_id FROM memberships WHERE org_id = '${ORG_B}';`
+  );
+  return {
+    name: "cannot see another org's membership roster",
```
barryroodt, line 308: Grades `memberships`, but the prompt only mentions documents. It's blocked-only too, so an agent that enables RLS on `memberships` with no policies scores the same as one that got it right. Every other negative here has a positive twin. Adding `editor can still see their own org's roster` makes it two-sided.

evals/build-rls-003-org-roles-permissions/EVAL.ts
```diff
@@ -1,227 +1,331 @@
...
+  ctx: ToolEvalContext,
+  sub: string,
+  body: string,
+  finish: 'COMMIT' | 'ROLLBACK' = 'COMMIT'
+): Promise<UserQueryResult> {
+  try {
+    const { rows } = await ctx.query(stripIndent`
+      BEGIN;
+      SET LOCAL ROLE authenticated;
+      SET LOCAL request.jwt.claim.sub = '${sub}';
```
barryroodt, line 70: `auth.jwt()` is always `{}` here. lite reads `auth.uid()` from `request.jwt.claim.sub` but `auth.jwt()` from the JSON `request.jwt.claims`, which we never set, so an agent using `auth.jwt() ->> 'sub'` gets denied everywhere and scores wrong. One line:

```sql
SET LOCAL request.jwt.claims = '{"sub":"${sub}","role":"authenticated"}';
```

evals/build-rls-003-org-roles-permissions/EVAL.ts
```diff
@@ -1,227 +1,331 @@
...
+ * technique shown in RLS testing guide:
+ * (https://supabase.com/docs/guides/local-development/testing/overview).
+ *
+ * Returns errors (incl. RLS-blocked writes) as values instead of throwing.
+ */
+async function runAsUser(
+  ctx: ToolEvalContext,
+  sub: string,
+  body: string,
+  finish: 'COMMIT' | 'ROLLBACK' = 'COMMIT'
```
barryroodt: Defaults to `'COMMIT'` and `checkViewerCannotInsert` takes the default, so a failing agent gets a junk document committed into the scored project. Harmless today since nothing counts documents. Flip the default to `'ROLLBACK'`?

evals/build-rls-003-org-roles-permissions/EVAL.ts
```diff
@@ -1,227 +1,331 @@
...
+      UPDATE documents
+      SET body = 'editor changed admin document'
+      WHERE id = '10000000-0000-0000-0000-000000000001'
+      RETURNING id;
+    `,
+    'ROLLBACK'
+  );
+  return {
+    name: "editor cannot update another user's document",
+    passed: Boolean(result.error) || result.rows.length === 0,
```
barryroodt: `Boolean(result.error)` counts `relation does not exist` and `current transaction is aborted` as "RLS blocked it". Tighten to `/row-level security|permission denied/i`, or `code === '42501'`.

barryroodt (commented):
Prompt rewrite reads like an actual person now, nice.

> Agents already pass this across the board, which is against our typical criteria for adding benchmarks, but maybe we're okay with that here.

Fine by me to make the exception here.

On hardening, two of the 14 checks measure less than they look like:

- **`RLS enabled on documents` can't fail anymore.** The seed now enables RLS (`project.sql:22`) and the check reads that same flag. Repoint it at the hole you seeded: assert no `SELECT` policy with `qual = 'true'` survives in `pg_policies`.
- **The roster check rewards nuking access** (inline).

Heads up that both of those change what's graded, so they'd need a results refresh. Your call whether that's worth it now or in a follow-up.

Also worth knowing: 7 of the 14 pass before the agent does anything. The seed turns RLS on and adds no write policies, so every "cannot insert / update / delete" is satisfied from the start.

evals/build-rls-003-org-roles-permissions/EVAL.ts
mattrossman: Confirmed and fixed in https://github.com/supabase/evals/commit/166c9f176be0eef0ffb2d84d484673471a94a459 by also setting `request.jwt.claims`, verified by running a correct `auth.jwt() ->> 'sub'` solution through the scorer which passed 9/15 before, 15/15 after.

evals/build-rls-003-org-roles-permissions/EVAL.ts
mattrossman: Added `editor can still see their own org's roster` plus a prompt line about org membership visibility in https://github.com/supabase/evals/commit/166c9f176be0eef0ffb2d84d484673471a94a459, and confirmed the deny-everything version now scores 14/15 instead of passing clean.

evals/build-rls-003-org-roles-permissions/EVAL.ts
mattrossman: Went one further in https://github.com/supabase/evals/commit/166c9f176be0eef0ffb2d84d484673471a94a459 and dropped the `finish` param entirely since no check needs `COMMIT`, which also kills the hidden ordering dependency where the viewer `SELECT` only saw 2 rows because it ran before the viewer `INSERT`.

evals/build-rls-003-org-roles-permissions/EVAL.ts
mattrossman: Switched to SQLSTATE `42501` in https://github.com/supabase/evals/commit/166c9f176be0eef0ffb2d84d484673471a94a459, which covers the RLS violation and the missing-grant case while excluding `42P01` and `25P02`, verified by renaming `documents` away and watching the negative checks fail instead of pass. I preferred the code over regex.

mattrossman:
@barryroodt Refreshed results are in https://github.com/supabase/evals/commit/9ebb571, still all passing.

> RLS enabled on documents can't fail anymore

I'm okay leaving this as a sanity check, because even though it's seeded for the footgun policy, I don't want agents doing any weird workarounds that might disable it (for example dropping + recreating the table to rename). I skipped the `pg_policies` swap, because the overly permissive seeded policy is already caught by `viewer sees only documents in their org` check.

> Also worth knowing: 7 of the 14 pass before the agent does anything

I think this is fine as long as our negative checks have positive twins that fail at the start. The initially passing checks exist to make sure the agent doesn't accidentally undo any desirable initial behaviors.

barryroodt (approved):
Looks good!

PR #165, opened 2026-08-04
feat(web): add social and SEO meta tags

seanoliver:
Links to supabase.com/evals unfurl with no preview card, and search results have no description.

CleanShot 2026-08-04 at 12 00 10

This PR adds description, canonical, Open Graph and `twitter:card` tags to `apps/web/index.html`, plus an evergreen `og.png` for the card.

**Rationale:**

- Tags are static in the shell because unfurlers do not execute JS. Nothing here needs SSR.
- URLs are absolute. Crawlers do not resolve a relative `og:image`.
- `twitter:card` is the only `twitter:` tag needed. X falls back to the `og:` equivalents for title, description and image.
- `description` and `og:description` carry the same copy, since neither falls back to the other. Copy is Matt's, at 96 characters so search results do not truncate it.
- The original card reused the launch post's image, which read "Introducing Supabase Evals" on a page that is now permanent. Swapped for a plain "Supabase Evals" card.

Built locally: tags land in `dist/index.html` and the image at `dist/og.png`, which serves at `/evals/og.png` once deployed.

Fixes https://linear.app/supabase/issue/GROWTH-1091/add-social-and-seo-meta-tags-to-supabasecomevals


raulb (approved):
@seanoliver Thank you! . Looks good to me though the image should probably be different? @SaxonF what do you think?

raulb:
> @seanoliver Thank you! . Looks good to me though the image should probably be different? @SaxonF what do you think?

Looks like @mattrossman already followed up [here](https://supabase.slack.com/archives/C0BAGJBL49E/p1785876840967319?thread_ts=1785870395.341699&cid=C0BAGJBL49E)

apps/web/index.html
```diff
@@ -5,6 +5,21 @@
     <link rel="icon" href="/favicon/favicon.ico" sizes="any" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Supabase Evals</title>
+    <meta
+      name="description"
+      content="We evaluate model experiments across the Supabase developer journey, from building and deploying to investigating and resolving production issues."
```
mattrossman: Might be a good opportunity to incorporate Terry's [copy feedback](https://supabase.slack.com/archives/C0BAGJBL49E/p1785853686428939?thread_ts=1785551136.746719&cid=C0BAGJBL49E) as I recognize this is the same tagline text.

mattrossman: Maybe something like this to reduce jargon?

```suggestion
      content="The open-source Supabase benchmark for AI coding agents."
```

mattrossman: Or a bit longer:

```suggestion
      content="See how agents perform on real Supabase tasks in our open-source benchmark for AI coding agents."
```

seanoliver: Great point. I actually really like the longer version for OG metadata specifically. It gives the reader more of a reason to click / call-to-action.

mattrossman (approved):

PR #168, opened 2026-08-04
feat: eval the Row Level Security guide in supabase.com/docs

czenko:
Closes DOCS-1259

## Problem

We point agents at the Row Level Security guide. Nobody has measured whether reading it is enough to produce secure policies.

The previous version of this eval could not answer that. It copied the guide's AI prompt panel into `PROMPT.md`, so the eleven rules that spell out the answer were the prompt itself.

## Solution

The guide is now the subject under test. The prompt is a normal feature request plus the guide's url. The eleven rules became the grading criteria.

The prompt is written as a vibe coder. It never says RLS, policy, security, role, tenant or test. An agent that only writes secure policies when the prompt says "security" has not been served by the guide.

The seed is two separate apps, so one access pattern cannot be applied to everything.

- Owner-private. `todos`.
- Shared through membership. `lists`, `list_members`, `list_items`.
- Public read with no client writes. `weather_stations`, `weather_readings`.

`using (true)` is correct on the weather feed and catastrophic on todos. `list_members` forces a `security definer` helper, because a policy that joins it is a join inside a policy, and a policy that queries itself raises `42P17`.

Nothing in the migration says who may read or write what.

The prompt does not ask for tests. Whether the agent arrives at pgTAP is part of the measurement.

## Rename

Both docs-prompt evals move into a series named for the documentation under test.

| From | To |
| --- | --- |
| `build-cli-004-quickstart-nextjs` | `build-docs-001-homepage-quickstart` |
| `build-cli-005-quickstart-rls` | `build-docs-002-rls-guide` |

Renaming breaks the gh-pages history series for 004, which is keyed by eval id.

## Testing

`pnpm typecheck` and `pnpm format:check` are clean. `pnpm eval:dry` confirms discovery and frontmatter for both ids. Results are refreshing in CI.

The scorer has not been run end to end. Local runs were blocked by another project's stack holding the Supabase ports.


evals/build-cli-005-quickstart-rls/EVAL.ts
```diff
@@ -0,0 +1,791 @@
...
+        ? anon.error.code === '42501'
+        : (anon.data?.length ?? 0) === 0,
+      notes: anon.error
+        ? describeError(anon.error)
+        : `${anon.data?.length ?? 0} rows`,
+    },
+  ];
+}
+
+async function checkDocumentWrites(
```
barryroodt: **Deny-all UPDATE and DELETE policies can pass this scorer.** These checks cover successful inserts and denied updates, but never a successful owner update or any delete. Could we prove allowed writes change the row, then prove foreign-user writes leave it unchanged?

evals/build-cli-005-quickstart-rls/EVAL.ts
```diff
@@ -0,0 +1,791 @@
...
+    name: 'columns the policies filter on are indexed',
+    passed: missing.length === 0,
+    notes:
+      missing.length === 0
+        ? undefined
+        : `no leading-column index on: ${missing.map((c) => `${c.table}.${c.column}`).join(', ')}`,
+  };
+}
+
+function checkPrivateHelperExists(helpers: HelperRow[]): CheckResult {
```
barryroodt: **The policy can call an unsafe helper and still pass.** `search_path_pinned` accepts any configured path, and the policy check can resolve a different helper. Could we require `search_path = ''` on the exact helper referenced by `team_documents`?

evals/build-cli-005-quickstart-rls/EVAL.ts
```diff
@@ -0,0 +1,791 @@
...
+  return {
+    name: `supabase test db runs at least ${MIN_PGTAP_TESTS} assertions and all pass`,
+    passed: result.ok && ran && failed === 0 && total >= MIN_PGTAP_TESTS,
+    notes: ran
+      ? `${total - failed} passed, ${failed} failed, exit ${result.exitCode}`
+      : `no test summary found; exit ${result.exitCode}; output: ${output.slice(0, 500)}`,
+  };
+}
+
+async function checkTestCoverage(
```
barryroodt: **Unrelated pgTAP tests can satisfy this check.** A scratch-table suite, or `lives_ok()` around a denied zero-row write, can meet this rubric. Could we bind coverage to the seeded tables and require allowed writes to change state?

evals/build-cli-005-quickstart-rls/EVAL.ts
```diff
@@ -0,0 +1,791 @@
...
+    },
+    {
+      name: 'user A cannot hand their own document to another user',
+      passed: docAStillOwned,
+      notes: docAStillOwned ? undefined : `rows now: ${JSON.stringify(rows)}`,
+    },
+  ];
+}
+
+async function checkTeamDocumentVisibility(
```
barryroodt: **Several fixture rules have no live probe.** The scorer never verifies member reads of `teams` or allowed writes to `team_documents`. Could we exercise those paths through the signed-in clients?

evals/build-cli-005-quickstart-rls/EVAL.ts
```diff
@@ -0,0 +1,791 @@
...
+    name: 'policies wrap auth.uid()/auth.jwt() in a select instead of calling per row',
+    passed: offenders.length === 0,
+    notes:
+      offenders.length === 0
+        ? undefined
+        : `unwrapped call in: ${offenders.map((p) => `${p.tablename}.${p.policyname}`).join(', ')}`,
+  };
+}
+
+async function checkPolicyColumnsIndexed(
```
barryroodt: **A non-btree or partial index can pass rule 5.** This query accepts any valid index with the column first. Could we require btree and reject partial indexes?

evals/build-cli-005-quickstart-rls/PROMPT.md
```diff
@@ -0,0 +1,82 @@
...
+   leads a primary key or unique index does not need another one.
+6. Avoid joins inside policies. Instead of joining the target table to the
+   source table, select the matching rows from the target table filtered by
+   `(select auth.uid())` and use `in` or `any` against that set.
+7. When a check needs to bypass RLS on a lookup table, such as team membership
+   or role checks, call a `security definer` function instead of joining.
+   Create it in a private, non-exposed schema, declare `set search_path = ''`
+   and schema-qualify every reference inside it, and verify `auth.uid()` in the
+   function body.
+8. Keep the equivalent filter in application queries even though policies act
```
barryroodt: **Rule 8 is currently unmeasured.** The fixture has no application-query surface, so an agent can ignore equivalent query filters. Could we add a check, or call this rule out as intentionally out of scope in the README?

evals/build-cli-005-quickstart-rls/EVAL.ts
```diff
@@ -0,0 +1,791 @@
...
+  'supabase_functions',
+  'supabase_migrations',
+  'information_schema',
+  'tests',
+]);
+
+/** Allow + deny for four operations across two roles is eight assertions. */
+const MIN_PGTAP_TESTS = 8;
+
+const scorer: LocalStackScorer = async (ctx) => {
```
barryroodt: **This scorer now carries four responsibilities in one 791-line file.** Catalog checks, fixture setup, access scenarios, and pgTAP inspection will all grow together. Could we split catalog and access modules now, keeping `EVAL.ts` as the phase orchestrator?

barryroodt (changes requested):
Thanks, this is a useful eval and the core shape looks good. `pnpm format:check` and `pnpm typecheck` pass. I found several false-pass paths that need closing before merge, and I’ve left the exact cases inline.

The PR also needs refreshed eval results. The handwritten solution proves one passing path. Real agent results show whether the eval discriminates.

`pnpm check` failed in one results-export test; attribution is unresolved.

evals/build-cli-005-quickstart-rls/PROMPT.md
czenko: After discussing the docs Product Manager, I've completely rewritten the prompt and the intent of the test.

evals/build-cli-005-quickstart-rls/EVAL.ts
czenko: Changed this so that the query joins pg_am and requires am.amname = 'btree' and
i.indpred IS NULL.

evals/build-cli-005-quickstart-rls/EVAL.ts
czenko: Probes added 🙏

evals/build-docs-002-rls-guide/access.ts
```diff
@@ -0,0 +1,523 @@
...
+    stripIndent`
+      INSERT INTO todos (user_id, title) VALUES
+        ('${fixtures.userAId}', '${fixtures.todoA}'),
+        ('${fixtures.userAId}', '${fixtures.todoAEdit}'),
+        ('${fixtures.userAId}', '${fixtures.todoADelete}'),
+        ('${fixtures.userBId}', '${fixtures.todoB}');
+
+      INSERT INTO lists (id, owner_id, name)
+        VALUES ('${fixtures.listId}', '${fixtures.userAId}', '${fixtures.listName}');
+      INSERT INTO list_members (list_id, user_id)
```
barryroodt, line 114: **An agent that enrols the list owner as a member via a trigger crashes the scorer to zero checks.** The insert has no `ON CONFLICT`, and that's what produced the current head result: one `scorer evaluated RLS policies and tests` record where `HEAD^` had 35. Could we add `ON CONFLICT (list_id, user_id) DO NOTHING` with a `BEGIN`/`COMMIT` around the block, then rebase and re-run the export rather than hand-merging the results file?

evals/build-docs-002-rls-guide/access.ts
```diff
@@ -0,0 +1,523 @@
...
+        VALUES ('${fixtures.listId}', '${fixtures.userAId}');
+      INSERT INTO list_items (list_id, author_id, title)
+        VALUES ('${fixtures.listId}', '${fixtures.userAId}', '${fixtures.listItem}');
```
barryroodt, line 125: **Owner-only plus author-only policies clear all six shared-list checks while sharing is completely broken, and a permissive cross-user write policy passes too.** `userA` is the owner, the only `list_members` row and the author of the one item, and the whole block runs seven SELECTs and a single INSERT. Could we add a `userC` who is only a member, then add outsider update and delete probes plus an anon read of `list_members`?

evals/build-docs-002-rls-guide/catalog.ts
```diff
@@ -0,0 +1,412 @@
...
+        : incomplete.length === 0
+          ? undefined
+          : `missing a clause: ${incomplete.map((p) => `${p.tablename}.${p.policyname}`).join(', ')}`,
+  };
+}
+
+// Postgres renders a wrapped call as `( SELECT auth.uid() AS uid)` and a bare
+// one as `auth.uid()`, so comparing the two counts catches per-row calls
+// without depending on how the agent spaced its SQL.
+export function checkAuthCallsWrapped(policies: PolicyRow[]): CheckResult {
```
barryroodt, line 279: **A policy that trusts `user_metadata` scores green, which is self-service admin over everyone's todos.** Neither fixture user has metadata set, so the admin clause evaluates false for both and every probe behaves as if the policy were correct. Could we have `clientB` set `{ admin: true }` on itself, refresh the session, and re-run the cross-user read?

evals/build-docs-002-rls-guide/catalog.ts
```diff
@@ -0,0 +1,412 @@
...
+    name: 'columns the policies filter on have a btree index',
+    passed: missing.length === 0,
+    notes:
+      missing.length === 0
+        ? undefined
+        : `no complete btree index leading with: ${missing.map((c) => `${c.table}.${c.column}`).join(', ')}`,
+  };
+}
+
+export function checkListItemsAvoidsJoin(policies: PolicyRow[]): CheckResult {
```
barryroodt: **Following the guide fails this check.** The guide's "Minimize joins" fix still selects `FROM` the membership table, only dropping the correlation to the source table, so `list_id in (select list_id from list_members where user_id = (select auth.uid()))` trips `/\bFROM\s+(public\.)?list_members\b/i`. Could we match the correlation the guide names, or drop the check and lean on the probes plus the 42P17 case?

evals/build-docs-002-rls-guide/tests.ts
```diff
@@ -0,0 +1,133 @@
...
+    return { name, passed: false, notes: 'no test files to review' };
+  }
+
+  const sources = await Promise.all(
+    files.map(async (file) => `-- ${file}\n${await ctx.readFile(file)}`)
+  );
+  const corpus = sources.join('\n\n').toLowerCase();
+
+  const untested = REQUIRED_TABLE_GROUPS.filter(
+    (group) => !group.some((table) => corpus.includes(table))
```
barryroodt: **One comment plus eight `select ok(true)` clears every mechanical gate here.** Coverage is a lowercased substring search over raw file text with `.some()` over grouped tables, so three names out of six suffice and comments count. Could we require each seeded table individually and match against the assertion descriptions `pg_prove` echoes instead?

evals/build-docs-002-rls-guide/catalog.ts
```diff
@@ -0,0 +1,412 @@
...
+};
+
+export async function loadTableState(
+  ctx: LocalStackEvalContext
+): Promise<TableState[]> {
+  const { rows } = await ctx.query(stripIndent`
+    SELECT c.relname, c.relrowsecurity, c.relforcerowsecurity
+    FROM pg_class c
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    WHERE n.nspname = 'public' AND c.relkind = 'r'
```
barryroodt: **A view over `todos` without `security_invoker = true` bypasses RLS entirely and scores green.** `loadTableState` filters `relkind = 'r'` and every check ranges over the six seeded names, so anything else the agent leaves in `public` is outside the measurement. Could we let the RLS and grant assertions range over the exposed schema, keeping `SEEDED_TABLES` as the must-exist list?

barryroodt (changes requested):
Thanks, this is much better. Making the guide the subject and the prompt a plain feature request is what makes the eval mean something, and `checkTodoWrites` is now a properly strong assertion set: it proves the owner's update changed the row, the delete removed it, and re-reads as superuser instead of trusting the client error. Prior findings 1, 2, 5 and 7 are closed.

Two things block merge: the fixture crashes on a plausible correct solution, and the shared-list tables are the least tested surface in the eval, which is the opposite of what README:23 says they are. Both inline.

Add a third user who is a member of a list they do not own. This makes the eval test membership access instead of letting owner-only policies pass.

evals/build-docs-002-rls-guide/catalog.ts
czenko: Done. Thank you.

evals/build-docs-002-rls-guide/access.ts
czenko: Made the revision. Running the export. 😄

czenko:
@barryroodt Feedback applied here [2386cc3](https://github.com/supabase/evals/pull/168/commits/2386cc396a9f25b3de4da5f0bb4b594329156749) and eval results ran here: [c5d9588](https://github.com/supabase/evals/pull/168/commits/c5d9588b7d772a76a3b5d069bcdc87b3c2af4af4)

evals/build-docs-002-rls-guide/access.ts
```diff
@@ -0,0 +1,621 @@
...
+
+// user_metadata is writable by the user it belongs to, so a policy that trusts
+// a claim from it hands every user admin over everyone else's rows. Neither
+// fixture user has metadata by default, which makes such a policy look correct.
+export async function checkMetadataEscalation(
+  f: Fixtures
+): Promise<CheckResult> {
+  const name =
+    'a user cannot widen their own access by editing their user metadata';
+  const update = await f.clientB.auth.updateUser({ data: { admin: true } });
```
barryroodt: **The escalation probe catches one spelling on one table.** It sets `{ admin: true }` and re-reads only `todos`, so `using (owner_id = (select auth.uid()) or ((select auth.jwt()) -> 'user_metadata' ->> 'role') = 'admin')` on `lists` scores green, and `->>` on an absent key is NULL so no other spelling is reached. Two more gaps in the same 25 lines: `refreshSession()` at line 288 is unbound, and a failed refresh leaves the old token in place and the check green against the vulnerable policy; and `passed: !read.error && leaked.length === 0` at line 294 is true when the read returns nothing at all. Could we set the whole key set in one `updateUser`, bind the refresh and assert the claim landed in the token, then re-read `lists`, `list_items` and `list_members` alongside `todos` with `titles.includes(f.todoB)` as the positive control?

evals/build-docs-002-rls-guide/tests.ts
```diff
@@ -0,0 +1,148 @@
...
+  });
+
+  return { name, passed: verdict.passed, judgeNotes: verdict.notes };
+}
+
+function stripNonSql(sql: string): string {
+  return sql
+    .replace(/\/\*[\s\S]*?\*\//g, ' ')
+    .replace(/--[^\n]*/g, ' ')
+    .replace(/'(?:[^']|'')*'/g, ' ');
```
barryroodt: **The literal strip removes table names from the assertion families your own rubric accepts.** `policies_are('public','list_members', ARRAY[...])` and `table_privs_are('weather_stations','anon', ARRAY['SELECT'])` carry the table name as a quoted literal, so after `stripNonSql` the corpus contains neither and the gate returns `tests never reference: list_members` without ever calling the judge. Requiring all six individually was the right half of the fix. The other half is still open: the gate proves a name appears in surviving text, not that anything is asserted against it, so eight `ok(true)` calls plus scratch DDL naming the six tables still clears it. `checkPgTapSuitePasses` already captures the TAP stream. Could we require each table to appear in an `ok N - <description>` line pg_prove echoed, and drop the text scan?

evals/build-docs-002-rls-guide/catalog.ts
```diff
@@ -0,0 +1,401 @@
...
+          ]
+            .filter(Boolean)
+            .join('; '),
+  };
+}
+
+// A view runs as its owner unless security_invoker is set, so a view over a
+// protected table hands out everything the policies were meant to withhold.
+export function checkViewsRunAsInvoker(tables: TableState[]): CheckResult {
+  const views = tables.filter((t) => t.relkind === 'v' || t.relkind === 'm');
```
barryroodt: **This check cannot be satisfied for a materialized view, and rejects two of the four boolean spellings Postgres takes.** Line 155 selects `relkind === 'v' || relkind === 'm'`, but `security_invoker` is not a valid reloption on a matview, so an agent that creates one fails with no available fix. Line 160 accepts only `true|on`, so `WITH (security_invoker = 1)` or `= yes` reads as leaky. Worth checking the pgTAP case too, since the eval requires a pgTAP suite and pgTAP creates its own views when installed into `public`. Could we restrict this to `relkind = 'v'`, widen the alternation to `(true|on|yes|1)`, and anti-join `pg_depend` on `deptype = 'e'` so extension-owned relations drop out?

evals/build-docs-002-rls-guide/catalog.ts
```diff
@@ -0,0 +1,401 @@
...
+  ctx: LocalStackEvalContext
+): Promise<CheckResult[]> {
+  const { rows } = await ctx.query(stripIndent`
+    SELECT c.relname AS table_name, r.rolname, p.priv
+    FROM pg_class c
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    CROSS JOIN (VALUES ('anon'), ('authenticated')) AS r(rolname)
+    CROSS JOIN (VALUES ('INSERT'), ('UPDATE'), ('DELETE')) AS p(priv)
+    WHERE n.nspname = 'public'
+      AND c.relkind IN ('r', 'p', 'v', 'm')
```
barryroodt: **The broadening false-fails a correct solution that adds a read-only view.** The scan is `relkind IN ('r','p','v','m')`, and Supabase's default privileges hand `anon` INSERT, UPDATE and DELETE on every new relation in `public`, including views those privileges cannot act through. For ordinary tables the broadening is a sharpening and I would keep it. Could we add `AND c.relkind IN ('r','p')` so the scan covers only relations a write privilege can reach?

evals/build-docs-002-rls-guide/catalog.ts
```diff
@@ -0,0 +1,401 @@
...
+// measured too, rather than only the seeded names.
+export async function checkGrants(
+  ctx: LocalStackEvalContext
+): Promise<CheckResult[]> {
+  const { rows } = await ctx.query(stripIndent`
+    SELECT c.relname AS table_name, r.rolname, p.priv
+    FROM pg_class c
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    CROSS JOIN (VALUES ('anon'), ('authenticated')) AS r(rolname)
+    CROSS JOIN (VALUES ('INSERT'), ('UPDATE'), ('DELETE')) AS p(priv)
```
barryroodt: **The name claims more than the check measures.** `anon holds no write grant anywhere in the exposed schema` is proven by three privileges, so `REVOKE INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public FROM anon` scores green while anon keeps TRUNCATE and can still empty every table. Could we add `('TRUNCATE')` to the VALUES list, or narrow both names to say insert, update and delete?

evals/build-docs-002-rls-guide/access.ts
```diff
@@ -0,0 +1,621 @@
...
+    SELECT title FROM list_items
+    WHERE title IN ('${created}', '${f.listItem}', 'outsider-hijacked')
+  `);
+  const { rows: listRows } = await ctx.query(stripIndent`
+    SELECT name FROM lists WHERE id = '${f.listId}'
+  `);
+  const itemTitles = rows.map((row) => String(row.title));
+  const listIntact = listRows[0]?.name === f.listName;
+
+  const recursive = [
```
barryroodt, line 422: **A 42P17 recursion on the write path scores green.** This array collects errors from the bound SELECTs and now `anonMembership.error`, but the three outsider write results added in the same commit are discarded, so a policy that recurses only on UPDATE or DELETE leaves every collected error clean. Could we bind those three and add their `.error` values, and add a member write through `clientC` while you are there, since the write path has no legitimate-user coverage?

evals/build-docs-002-rls-guide/access.ts
```diff
@@ -0,0 +1,621 @@
...
+        ('${fixtures.userAId}', '${fixtures.todoAEdit}'),
+        ('${fixtures.userAId}', '${fixtures.todoADelete}'),
+        ('${fixtures.userBId}', '${fixtures.todoB}');
+
+      INSERT INTO lists (id, owner_id, name)
+        VALUES ('${fixtures.listId}', '${fixtures.userAId}', '${fixtures.listName}');
+      INSERT INTO list_members (list_id, user_id) VALUES
+        ('${fixtures.listId}', '${fixtures.userAId}'),
+        ('${fixtures.listId}', '${fixtures.userCId}')
+        ON CONFLICT (list_id, user_id) DO NOTHING;
```
barryroodt: **`ON CONFLICT (list_id, user_id)` pins a schema the prompt invites the agent to change.** It requires a unique constraint matching that inference exactly, so an agent who models membership as a surrogate id plus `unique (list_id, user_id, role)` gets `there is no unique or exclusion constraint matching the ON CONFLICT specification`, which is the same run-collapsing throw as before in a new shape. Could we use a constraint-agnostic seed instead: `INSERT ... SELECT ... FROM (VALUES ...) v WHERE NOT EXISTS (SELECT 1 FROM list_members m WHERE m.list_id = v.list_id AND m.user_id = v.user_id)`?

barryroodt (changes requested):
Strong response. Items 2, 4 and 6 are closed, and deleting `checkListItemsAvoidsJoin` instead of patching the regex was the right call. The results file now carries a complete run: 39 checks, 34 passing, five failures that all describe real agent defects.

What's left, all inline:

- **Item 3 is open.** The escalation probe sets one metadata key and re-reads one table.
- **Item 5 is half done.** Requiring all six tables landed. The gate still matches file text, and the new literal strip breaks `policies_are(...)`.
- **Two of the new checks false-fail.** Materialized views cannot carry `security_invoker`, and the grant scan now covers views.
- **Item 1's `ON CONFLICT` pins a schema the prompt invites the agent to change.**
- **The rebase is still outstanding.** Main gained two more commits that rewrite the results file, so re-running the refresh on the new base beats hand-resolving it.

Minor, not threaded: `outsider-hijacked` isn't run-scoped, `execSql` still throws past `setupFixtures`, `REQUIRED_TABLES` duplicates `SEEDED_TABLES`, three renamed checks say "the exposed schema" but query `public`, and the README still describes six tables.

czenko:
So appreciate this, @barryroodt 🙏

czenko:
@barryroodt I so appreciate your thorough reviews. All five are applied in c09a674.

**Escalation probe.** One `updateUser` sets six spellings of the claim. The refresh is bound, and the check fails if the claim never reached the token. It re-reads `lists`, `list_items` and `list_members` alongside `todos`, with the user's own todo as the positive control.

**Coverage gate.** The text scan is gone, `stripNonSql` with it. Coverage now comes from the `ok N` lines pg_prove echoes.

**The two false-fails.** The invoker check is restricted to `relkind = 'v'` and takes all four boolean spellings. 

**Grant name.** Took the first option and added `TRUNCATE` to the privilege list, rather than narrowing the names.

**ON CONFLICT.** Replaced with the constraint-agnostic seed you suggested. The weather feed restore lost its `ON CONFLICT (id)` for the same reason.

**Minors, all applied.**

* `outsider-hijacked` and the todos `hijacked` marker are run-scoped.
* `execSql` returns instead of throwing. A seed that fails now costs the probes that need it, not every check.
* `REQUIRED_TABLES` is gone. `SEEDED_TABLES` is the one list.
* Three check names say `public` schema.
* README covers the exposed-schema scope and how coverage is measured.

evals/build-docs-002-rls-guide/catalog.ts
```diff
@@ -0,0 +1,409 @@
...
+          ]
+            .filter(Boolean)
+            .join('; '),
+  };
+}
+
+// A view runs as its owner unless security_invoker is set, so a view over a
+// protected table hands out everything the policies were meant to withhold.
+export function checkViewsRunAsInvoker(tables: TableState[]): CheckResult {
+  const views = tables.filter((t) => t.relkind === 'v');
```
barryroodt, line 163: **A materialized view over `todos` now scores a clean green.** Narrowing this to `relkind = 'v'` was my ask and it was the wrong ask: `loadTableState` still selects `'m'` at line 82, `checkRlsEnabled` filters to `'r','p'` at 137, and `checkGrants` narrowed to `'r','p'` at 355, so nothing reads matviews at all. RLS never applies to a matview read and `security_invoker` is not accepted on one, so `create materialized view public.all_todos as select * from todos` hands every row to `anon` while this check reports `no views created`. Failing every matview outright is defensible where the old behaviour was not, because a matview is a choice the agent made rather than something Postgres makes unachievable. Could we partition on `relkind`, require the reloption on `'v'`, and fail `'m'` by name, with a `SELECT` grant probe on matviews as the backstop?

evals/build-docs-002-rls-guide/tests.ts
```diff
@@ -0,0 +1,149 @@
...
+): Promise<CheckResult> {
+  const name =
+    'tests assert allow and deny per operation on the seeded tables, for anon and authenticated';
+  const files = await listTestFiles(ctx);
+  if (files.length === 0) {
+    return { name, passed: false, notes: 'no test files to review' };
+  }
+
+  const reported = assertions.join('\n').toLowerCase();
+  const untested = SEEDED_TABLES.filter(
```
barryroodt: **The gate is too strict and too weak at once, so it cannot be tuned.** Too strict: a suite of `lives_ok(...)`, `throws_ok(..., '42501')` and `results_eq(...)` written without description arguments emits bare `ok N` lines, so `reported` is empty, all six tables land in `untested`, and this returns false before the judge is called. Too weak: one `ok(true, 'todos lists list_members list_items weather_stations weather_readings')` puts all six names in the joined text and clears it. Tightening one direction worsens the other. My "match the descriptions" suggestion caused this, and the honest fix is to drop the mechanical prefilter, since the judge already reads the file sources and its rubric is strictly stronger than a name match. Could we delete the prefilter and add the six tables to the rubric as an explicit per-table checklist?

evals/build-docs-002-rls-guide/access.ts
```diff
@@ -0,0 +1,717 @@
...
+  await client.from('weather_readings').delete().eq('id', f.readingId);
+
+  const { rows } = await ctx.query(stripIndent`
+    SELECT id::text AS id, conditions
+    FROM weather_readings
+    WHERE id = '${f.readingId}' OR conditions = '${marker}'
+  `);
+  const seeded = rows.find((row) => row.id === f.readingId);
+
+  const restore = await execSql(
```
barryroodt, line 652: **A failed restore is scored against the next role's policies.** Two changes landed here: the DELETE gained `OR id = '<readingId>'` so it now destroys the control row the old statement spared, and the INSERT lost `ON CONFLICT (id) DO UPDATE`. There is no `BEGIN` here, unlike the seed, so under `ON_ERROR_STOP=1` the DELETE autocommits on its own. `attemptFeedWrites` runs twice, anon then authenticated, and the second call reads whatever the first call's failed restore left behind, so a correct authenticated policy scores red. `restoreError` reaches only `describeFeedWrites` and is excluded from `isBlocked`, so no check fails on it. Could we wrap it in `BEGIN`/`COMMIT`, delete only the marker rows, use an `UPDATE` followed by a `WHERE NOT EXISTS` insert, and fold `restoreError` into a check of its own?

evals/build-docs-002-rls-guide/tests.ts
```diff
@@ -0,0 +1,149 @@
...
+  };
+}
+
+// The counts alone are not enough: a suite whose plan does not match the
+// assertions it ran reports `Tests: 6 Failed: 0` and still exits non-zero with
+// `Result: FAIL`, so the exit status is what decides.
+export async function runPgTapSuite(
+  ctx: LocalStackEvalContext
+): Promise<PgTapRun> {
+  // Without --debug, the test results the coverage check reads never reach us.
```
barryroodt: **`--debug` can red a fully passing suite.** `failed` is parsed from `stdout + stderr`, and `--debug` fills stderr with CLI logging and Postgres wire frames, so any debug line matching `/Failed:\s*(\d+)/i` sets `failed` non-zero at line 60. It cannot produce a wrong green, only a wrong red, and it also crowds the notes meant to explain a real failure. Nothing in the review asked for this. Could we drop `--debug`, keep the stdout and stderr split you correctly introduced, and parse `total` from `stdout` anchored on `/Files=\d+,\s*Tests=(\d+)/`?

evals/build-docs-002-rls-guide/access.ts
```diff
@@ -0,0 +1,717 @@
...
+          : undefined,
+    },
+    {
+      name: "a signed-in user cannot delete another user's todo",
+      passed: present(f.todoB),
+      notes: present(f.todoB) ? undefined : "the other user's todo was deleted",
+    },
+  ];
+}
+
```
barryroodt, line 288: **Enumerating six key names cannot close this class.** Two concrete misses in the current list: `(auth.jwt() -> 'user_metadata' -> 'claims' ->> 'admin')::boolean` sees no `claims` object because the probe writes only flat keys, and `auth.jwt() -> 'user_metadata' ->> 'role' = 'service_role'` is the right key with a value the probe does not set. Any list is incomplete by construction. Could we add a static companion using the existing `expressionsOf(policies)` that fails any policy whose `qual` or `with_check` mentions `user_metadata` at all, and keep the probe as the behavioural proof? There is no legitimate reason for a policy to read a field its own subject can write, so that rule carries no false-fail risk.

barryroodt (changes requested):
Almost there, thank you for your patience and for the continuous updates :). Everything I asked for landed, and two of the things I asked for were wrong.

Closed: threads 4, 5, 6 and 7, all four asks on thread 1, all six minors, and the rebase.

Three blockers. Two are inline:

1. Restricting the invoker check to `relkind = 'v'` was my ask, and it left materialized views measured by nothing.
2. Matching table names against pg_prove descriptions was also my ask. It rejects a correct suite written without descriptions, and `ok(true, '<six table names>')` still clears the gate.

The third is in `apps/web/src/data/regression-eval-results.json`, on a line outside the diff so I cannot thread it. The refresh reverted the eval rename: the record is keyed `build-cli-004-quickstart-nextjs` again, byte-matching main's copy, so `build-docs-001-homepage-quickstart` has no row and the file points at a deleted `PROMPT.md`. Re-running the export will not recover it, because the merge in `export-results.ts:278` is keyed on `experiment::eval`, so the new key is written and the stale record survives beside it. Needs three fields re-keyed to what `c5d9588` already had, plus the `results/claude-code-sonnet-5/` artifact renamed so the next export does not resurrect the orphan.

Three more inline: the weather-feed restore can lose the control row and score it against the next role, `--debug` can red a passing suite, and the escalation probe cannot close its class by listing key names.

czenko:
@barryroodt Appreciate your patience. 

All six are in 05d757a.

**Materialized views.** The invoker check partitions on `relkind`. A view must carry `security_invoker`, a matview fails by name, and a new check proves neither client role holds `SELECT` on one.

**Coverage gate.** The prefilter is deleted. The six tables are a per-table checklist in the judge's rubric instead, since the judge reads the sources anyway.

**`--debug`.** Dropped. `total` and `failed` come from pg_prove's summary line, anchored on `Files=\d+,\s*Tests=(\d+)`, read from stdout with the split you asked for.

**Feed restore.** One `BEGIN`/`COMMIT`. It deletes only the marker rows and repairs the control row with an `UPDATE` plus a `WHERE NOT EXISTS` insert. `restoreError` fails a check of its own now.

**Escalation probe.** A static check fails any policy whose `qual` or `with_check` mentions `user_metadata` at all. The probe stays as the live proof.

**Results key.** Re-keyed `eval`, `sourcePath` and `promptSourcePath` to what `c5d9588` had, then re-sorted the file the way the export does. Nothing to rename on the artifact side: `results/*/` is gitignored and the workflow rebuilds it per run from that run's own artifacts, so a stale copy is a local delete.


**About merging:**

- Since this has gone through several rounds of feedback and it seems our time zones make the turnaround a day each time, I was wondering if, in this final round, you can  push changes directly to this branch. I would appreciate getting this in. 🙏 ✨ 
- We have scoped ten more Evals of this nature. I'm wondering if there is something you would like me to do to help reduce review time (like a checklist I can follow or contextual resources) before coming to you for a review. Perhaps, as a best practice, always including a fixture pair (one red, one green). I can also test a fix before asking for review.

barryroodt (approved):
Approving. Thanks for the persistence 😄 and nice work on the matview partition!

On pushing directly, happy to do so for small things, especially if it can save everyone some time.

Your fixture idea is the right one. I'd suggest committing a couple of example solutions next to the eval, then run the scorer against them, so a run tells you whether the checks behave.

### Two kinds:

- **A green one.** A complete solution you believe is correct and secure, covering everything the eval scores, so for this one that means the pgTAP tests as well as the policies. Run the scorer. Every check should pass. If one fails, look at the solution and the check before the eval ships, since either could be the thing that's wrong.
- **A few bad ones.** Each with one deliberate flaw, say `using (true)` on `todos`. Write down which checks you expect to fail, then run it.

Don't aim for exactly one failing check per bad fixture. A real flaw usually trips several, and that's fine as long as what you got matches the list you wrote down.

### Some things worth considering as you write the checks:

- For each check, decide what should happen when the thing it looks at isn't there at all.
  - "RLS is enabled on `todos`" should fail if `todos` is missing. 
  - "No client role can read a materialized view" should pass if the agent made no matview. 
- Query the schema for the object kinds the eval means to cover, instead of hardcoding a list of table names. Then agent-added objects in that scope get measured too.

Might be worth putting this in `CONTRIBUTING.md` so it covers future evals. Happy to do so.

czenko:
Great. I can draft this feedback in CONTRIBUTING for you today. 😄

PR #178, opened 2026-08-07
fix(core): normalize claude-code MCP tool names to bare endpoint

Rodriguespn:
## What

The claude-code parser recorded MCP tool calls' `originalName` (the scorer-facing `endpoint`) as Claude Code's raw `mcp__<server>__<tool>` (e.g. `mcp__supabase-mcp__query_logs`), while the Codex parser records the bare `query_logs`. So a scorer doing `tc.endpoint === "query_logs"` matches Codex but **never** matches claude-code — tool-selection evals false-fail on claude-code for behavior that actually happened.

This strips the `mcp__<server>__` prefix in the claude-code parser so the endpoint is agent-agnostic, matching Codex. The raw name is still passed to `normalizeToolName`, so `name` stays `tool_use`.

## Why it's in the parser

Each agent parser owns normalizing its own tool names; Codex already emits the bare MCP tool name, claude-code was the outlier.

## Verification

- End-to-end: [`investigate-logs-002-query-logs-tool-selection`](https://github.com/supabase/evals/pull/100/changes#diff-a12cc4632ff0069bbe45f27140221aa5b121f78ca7758868977f338d082362deR4) with `claude-code-sonnet-5` against the mcp#333 preview build went **FAIL (1/2) → PASS (2/2)**; Showing a false negative (agent was called the tool but the eval check ignored it)

### DIFF
--- a/packages/core/src/agents/codex/parser.ts
+++ b/packages/core/src/agents/codex/parser.ts
@@ -24,10 +24,7 @@ import type {
   ToolCall,
   TranscriptEvent,
 } from '../../transcript/types.js';
-import type {
-  AgentTranscriptParser,
-  ParseContext,
-} from '../../parsers/types.js';
+import type { AgentTranscriptParser } from '../../parsers/types.js';
 import {
   normalizeToolName,
   type AgentToolMap,
@@ -133,10 +130,7 @@ function loadedSkillsFromCodexCall(
   return [];
 }
 
-function itemToEvents(
-  item: Record<string, unknown>,
-  soleServer?: string
-): TranscriptEvent[] {
+function itemToEvents(item: Record<string, unknown>): TranscriptEvent[] {
   const id = str(item.id) ?? '';
   const itemType = str(item.type);
 
@@ -182,11 +176,8 @@ function itemToEvents(
       // Shape not pinned across versions — be defensive about field names and
       // treat a missing status as unknown (not success). `item.tool` is the
       // bare tool name; `item.server` names the MCP server when present.
-      const bare =
-        str(item.tool) ?? str(item.name) ?? str(item.server) ?? 'mcp_tool_call';
-      // Prefer the explicit server field; fall back to the sole configured MCP
-      // server when the (unpinned) shape omits it.
-      const server = str(item.server) ?? soleServer;
+      const bare = str(item.tool) ?? str(item.name) ?? 'mcp_tool_call';
+      const server = str(item.server);
       return toolCallPair(
         id,
         bare,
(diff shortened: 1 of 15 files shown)


packages/core/src/agents/codex/parser.ts
```diff
@@ -169,15 +180,23 @@ function itemToEvents(item: Record<string, unknown>): TranscriptEvent[] {
     }
     case 'mcp_tool_call': {
       // Shape not pinned across versions — be defensive about field names and
-      // treat a missing status as unknown (not success).
-      const tool =
+      // treat a missing status as unknown (not success). `item.tool` is the
+      // bare tool name; `item.server` names the MCP server when present.
+      const bare =
         str(item.tool) ?? str(item.name) ?? str(item.server) ?? 'mcp_tool_call';
```
mattrossman: I know it's not new in this PR, but this fallback chain looks strange. I wouldn't expect the tool name to fall back to server name, maybe better to keep it undefined or generic fallback text if it's truly something that can go missing (Codex [source](https://github.com/openai/codex/blob/c18e9f478bc940ef1ef8e1c426364c0fe3d86b73/codex-rs/exec/src/exec_events.rs#L283-L290) seems to indicate it's not optional)

mattrossman:
Sanity checked running all 3 harnesses, `toolName` and related fields appear to be populating as expected! Claude examples below.

CleanShot 2026-08-11 at 11 45 19@2x 

CleanShot 2026-08-11 at 11 46 28@2x

packages/core/src/agents/codex/parser.ts
```diff
@@ -169,15 +180,23 @@ function itemToEvents(item: Record<string, unknown>): TranscriptEvent[] {
...
+      // Prefer the explicit server field; fall back to the sole configured MCP
+      // server when the (unpinned) shape omits it.
+      const server = str(item.server) ?? soleServer;
```
mattrossman: I think we can remove the concept of "soleServer" as a fallback entirely, seems like dead code given the version of Codex we're on. I'd assume it will always tell us which server an MCP tool call was associated with, otherwise I'd want it fail validation in some visible way so we can stay aware of parser behaviors.

mattrossman (approved):
Approving to unblock, see notes about about Codex fallback cleanup opportunities

Rodriguespn:
Thx for the comments @mattrossman. Cleaned Codex fallback logic and sanity ran codex gpt5.4-mini over `investigate-logs-001-top-error-function`. All good. Merging this 🚢  

image

PR #188, opened 2026-08-11
fix: keep LLM credentials out of the eval scorer environment

mattrossman:
Reduces secret exposure risk and fixes the (unused) eval where it was a risk.

- `vite build` and `vitest`scorer subprocesses no longer inherit CI's `process.env`, so agent can't accidentally print sensitive keys into result `notes`
- New regression test to verifies subprocesses can't see keys
- Scored projects now get `VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` from the harness, to fix `build-frontend-001-todos-app` failing with `supabaseUrl is required`
- Linking `node_modules` into the workspace now replaces whatever is there (previously, if agent ran `npm install` it'd populate `node_modules` and then block the symlink step and break `vite` deps)

Sanity checked `build-frontend-001-todos-app` runs + passes w/ the changes, plus a couple unrelated evals (`build-rls-002-own-todos-client` and `investigate-db-001-table-row-counts`) to check for regressions.

Closes AI-1004

### DIFF
--- a/apps/framework/harness/project-runner.ts
+++ b/apps/framework/harness/project-runner.ts
@@ -12,6 +12,13 @@ import type { CommandResult, VitestResult } from '@supabase-evals/core';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ROOT = join(__dirname, '..');
+const PROJECT_DB_URL = 'http://supabase-evals.local';
+const PROJECT_DB_ANON_KEY = 'supabase-evals-anon-key';
+const PROJECT_DB_JWT_SECRET = 'supabase-evals-dev-secret';
+const PROJECT_ENV = {
+  VITE_SUPABASE_URL: PROJECT_DB_URL,
+  VITE_SUPABASE_ANON_KEY: PROJECT_DB_ANON_KEY,
+};
 
 // Vite/vitest resolve their own package (and the workspace's deps, e.g. react)
 // by walking up from the workspace looking for a node_modules dir. Workspaces
@@ -26,7 +33,8 @@ export async function viteBuild(workspace: string): Promise<CommandResult> {
   return runNodeBin(
     join(ROOT, 'node_modules', 'vite', 'bin', 'vite.js'),
     ['build'],
-    workspace
+    workspace,
+    PROJECT_ENV
   );
 }
 
@@ -63,7 +71,7 @@ export async function vitestRun(workspace: string): Promise<VitestResult> {
       `--outputFile=${reportPath}`,
     ],
     workspace,
-    { SUPABASE_EVALS_WORKSPACE: workspace }
+    { ...PROJECT_ENV, SUPABASE_EVALS_WORKSPACE: workspace }
   );
   const parsed = existsSync(reportPath)
     ? parseVitestReport(reportPath)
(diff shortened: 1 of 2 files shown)


apps/framework/harness/project-runner.ts
```diff
@@ -157,7 +169,7 @@ async function runNodeBin(
   return new Promise((resolve) => {
     const child = spawn(process.execPath, [bin, ...args], {
       cwd,
-      env: { ...process.env, ...env },
+      env,
```
barryroodt, line 172: This flipped the fourth parameter's meaning from "extras merged over `process.env`" to "the child's whole environment", while keeping the name `env` and the `= {}` default. The invariant now lives at both call sites, so a future `runNodeBin(bin, args, cwd)` gets a bare env with no signal anything's wrong.

Non-blocking, but building it here deletes `PROJECT_ENV` and two of the four hunks in this file:

```ts
const PROJECT_CHILD_ENV = {
  VITE_SUPABASE_URL: PROJECT_DB_URL,
  VITE_SUPABASE_ANON_KEY: PROJECT_DB_ANON_KEY,
};
// ...
env: { ...PROJECT_CHILD_ENV, ...extraEnv },
```

`viteBuild` then goes back to its three-argument call, and `vitestRun` to `{ SUPABASE_EVALS_WORKSPACE: workspace }`.

apps/framework/scripts/smoke-framework.ts
```diff
@@ -475,20 +475,57 @@ async function smokeFrontendBuildTooling() {
...
-  writeFileSync(join(workspace, 'src', 'App.tsx'), GOOD_FRONTEND_APP);
+  writeFileSync(
+    join(workspace, 'tests', 'environment.test.ts'),
+    [
+      "import { expect, test } from 'vitest';",
+      '',
+      "test('does not inherit LLM credentials', () => {",
+      '  expect(process.env.ANTHROPIC_API_KEY).toBeUndefined();',
+      '  expect(process.env.OPENAI_API_KEY).toBeUndefined();',
+      '  expect(process.env.AI_GATEWAY_API_KEY).toBeUndefined();',
```
barryroodt, line 503: These three names are the only thing the fixture samples, so a revert that spreads `process.env` and filters out just these keys keeps `build.ok` and `vitest.ok` true while `GITHUB_TOKEN` and the rest still reach the child. Same goes for the `vite.config.ts` guard above.

Non-blocking: `expect(Object.keys(process.env).sort()).toEqual([...])` closes it, and it's easiest once the allowlist has one home (see my comment on `runNodeBin`).

barryroodt (approved):
Looks good! I put the narrowed env in front of the actual installed vite + vitest, and neither misses a thing from the parent. PATH's a non-issue too: esbuild resolves its native binary through `require.resolve`.

Left two non-blocking suggestions inline.

apps/framework/harness/project-runner.ts
mattrossman: I'm okay leaving as is with the `env` being an explicit thing callers provide, erring on the side of failing closed.

apps/framework/scripts/smoke-framework.ts
mattrossman: I took a middle ground in https://github.com/supabase/evals/pull/188/changes/c1882cae0ae0393ea1397f54a933d91e368d6806 where I assert that `PATH` doesn't leak in, when it'd otherwise be visible on `process.env`.

I tried your suggested strict equality check, however vite sets `NODE_ENV` and MacOS injects a `__CF_USER_TEXT_ENCODING` even with our controlled env on the spawned subprocess, so it's not as straightforward to ensure only those 3 keys were set.

PR #190, opened 2026-08-12
chore(evals): promote debugging skills evals to benchmark suite

Rodriguespn:
Promotes the existing debugging-skill scenarios to the `benchmark` suite so they surface on the [leaderboard](https://supabase.com/evals?group=eval) for the debugging-skills blog post and demo, adds one harder scenario that actually distinguishes skill from no-skill, and bumps the skill submodule to the version that ships the debugging workflow.

## Changes

**Promoted `regression` → `benchmark`** (added in #79, marked regression in #94). `suite` is a single enum, so this moves them into the benchmark + no-skills experiment matrix that feeds the leaderboard:

- `investigate-functions-001-546-resource-limit`
- `resolve-storage-001-upsert-missing-update-policy`
- `resolve-dataapi-002-update-zero-rows-affected`

**New eval — `resolve-performance-002-rls-auth-initplan`.** A large RLS-enabled `documents` table whose owner policy calls `auth.uid()` once *per row*; the fix is to wrap it in a scalar subquery — `(select auth.uid())` — so Postgres evaluates it once as an InitPlan. Scored deterministically (ownership preserved via `supabase-js`, `pg_policies.qual` wraps the auth call, RLS still on) plus a diagnosis judge. Grounded in [FDBKIN-32777](https://linear.app/supabase/issue/FDBKIN-32777) / [FDBKIN-25150](https://linear.app/supabase/issue/FDBKIN-25150) (RLS performance, "Churn risk").

**agent-skills submodule → v0.1.8.** Adds the `## Debugging` section to the `supabase` skill (fetch the Monitoring & Debugging docs before diagnosing). The benchmark was previously pinned to a pre-debugging skill — a large part of why the promoted three show no skill delta.

## Local results (both models, ±skills)

| eval | sonnet-5 | sonnet-5 no-skills | codex-mini | codex-mini no-skills |
|---|---|---|---|---|
| investigate-functions-001-546-resource-limit | ✅ | ✅ | ✅ | ✅ |
| resolve-storage-001-upsert-missing-update-policy | ✅ | ✅ | ✅ | ✅ |
| resolve-dataapi-002-update-zero-rows-affected | ✅ | ✅ | ✅ | ✅ |
| **resolve-performance-002-rls-auth-initplan** | ✅ 5/5 | ✅ 5/5 | ✅ 5/5 | ❌ **3/5** |

The promoted three are at ceiling — they map 1:1 onto the skill's own checklist, so frontier models pass with and without it. The new initplan eval is the first in this set to show a delta: codex-mini **without** the skill only adds an index and misses the InitPlan rewrite; **with** it, it applies `(select auth.uid())`. (`pnpm typecheck` and `biome check` clean; official roster numbers refreshed by CI.)

## Scenarios investigated but not shipped

Two realistic, high-volume debugging scenarios from Linear "Feedback Intake" + Slack were built and tested but left out, with rationale for the blog / skill authors:

- **PostgREST schema cache / PGRST204** — at ceiling (every model already knows PGRST204 → reload the cache) and not reproducible live in the harness (disabling PostgREST's `pgrst_ddl_watch` auto-reload needs superuser ownership migrations lack; confirms #79's skip).
- **Realtime `postgres_changes` + `REPLICA IDENTITY FULL`** — genuinely hard but a run-to-run coin-flip on the replica-identity insight (independent of the skill), and the skill has no realtime content to teach it.

## CI

Uses the `run-evals-changed` label to refresh only the changed evals (the promoted three + the new one) across the benchmark + no-skills roster. The full `run-evals` matrix now exceeds GitHub's 256-job cap (25 benchmark evals × 10 experiments + regression), so the whole-matrix refresh needs chunking as a separate follow-up.

Ref AI-1067

### DIFF
--- a/apps/web/src/data/eval-results.json
+++ b/apps/web/src/data/eval-results.json
@@ -1685,17 +1685,17 @@
       {
         "name": "identified video-thumbnails and the 546 resource-limit pattern",
         "passed": true,
-        "judgeNotes": "Identified video-thumbnails as the affected function and correctly described failures as HTTP 546 due to CPU/resource limit, not 500/503. Also treated welcome-email 500 as unrelated."
+        "judgeNotes": "Identified `video-thumbnails` as the affected function and correctly described failures as HTTP 546 due to CPU/resource limit, while treating `welcome-email` 500 as unrelated."
       },
       {
         "name": "attributed the 546s to CPU time exhaustion",
         "passed": true,
-        "judgeNotes": "The assistant explicitly attributes the 546 failures to CPU time exhaustion, citing `shutdown (reason: CPUTime, cpu_time_used: 2000ms, cpu_time_limit: 2000ms) → status 546`, and distinguishes it from application exceptions and unrelated welcome-email error."
+        "judgeNotes": "The assistant specifically attributes the HTTP 546 failures to CPU time exhaustion, citing log evidence: shutdown reason CPUTime and cpu_time_used/cpu_time_limit at 2000ms/2000ms. It does not blame memory, wall-clock time, code exceptions, or the unrelated welcome-email error."
       },
       {
         "name": "recommended reducing/offloading CPU work as the fix",
         "passed": true,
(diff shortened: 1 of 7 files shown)


Rodriguespn:
Closing this as the product team will own this - https://linear.app/supabase/project/observability-on-auto-pilot-cfa11f2cf529/activity#project-update-943e4cab&comment-f372b490

PR #114, opened 2026-07-23
feat: add Vercel Sandbox eval runner (AI-912 spike)

Rodriguespn:
## Check this [thread](https://supabase.slack.com/archives/C0BAGJBL49E/p1784919681390609) before doing any change to this PR. We might want to explore Vercel workflows to couple with sandboxes

Dispatch eval runs to Vercel Sandbox instead of GitHub Actions matrix jobs. 

Two pieces:

1. `packages/vercel-runner` reproduces the `eval-refresh.yml` pipeline end to end: matrix discovery, one isolated machine per `(experiment x eval)` pair running in parallel, the same `pnpm eval` command line, result collection, and the `publish-results` export step
3. `.github/workflows/eval-refresh-vercel.yml` is a manually-dispatched twin of `eval-refresh.yml` that runs the dispatcher from a single job. Same inputs, gating, and publish steps; Can be triggered by adding the `run-evals-sandbox` label to this PR.

## Validation

Local vs sandbox vs CI, same commit:

| experiment | eval | local | sandbox (local dispatch) | sandbox (CI dispatch) |
| --- | --- | --- | --- | --- |
| claude-sonnet-5 | investigate-db-001 | **PASS 3/3** | **PASS 3/3** | **PASS 3/3** ([run](https://github.com/supabase/evals/actions/runs/30037588926)) |
| claude-code-sonnet-5 | build-cli-001-bootstrap-app | — | — | **PASS 7/7** ([run](https://github.com/supabase/evals/actions/runs/30038495452)) |
| claude-code-sonnet-5-no-skills | build-cli-001-bootstrap-app | — | — | **PASS 7/7** ([run](https://github.com/supabase/evals/actions/runs/30038495452)) |
| codex-gpt-5.4-mini | build-cli-001-bootstrap-app | — | — | **PASS 7/7** ([run](https://github.com/supabase/evals/actions/runs/30038495452)) |
| codex-gpt-5.4-mini-no-skills | build-cli-001-bootstrap-app | — | — | **PASS 7/7** ([run](https://github.com/supabase/evals/actions/runs/30038495452)) |

## Limits (Pro plan)

[...]

PR #192, opened 2026-08-13
feat: move eval execution from GitHub Actions to Vercel Sandboxes

mattrossman:
Moves eval execution from GitHub Actions to [Vercel Sandboxes](https://vercel.com/docs/sandbox). The GitHub workflow remains for orchestrating sandboxes via a new `run-vercel-evals.ts` entrypoint. It uses [p-limit](https://www.npmjs.com/package/p-limit) and [p-retry](https://www.npmjs.com/package/p-retry) for concurrency limits and retry policies. Also considered [effect](https://effect.website/), but our use case isn't very complex to justify the runtime.

**Sandbox limits**

[As of Aug 5th](https://vercel.com/changelog/vercel-sandbox-now-supports-10-000-concurrent-sandboxes-and-5-000-vcpus-per-minute), our Pro plan supports 10,000 concurrent sandboxes and 5,000 vCPU allocated / min with dynamic allocation quotas that ramp up during sustained usage.

I'm starting with a default of **250** concurrent sandboxes for the workflow, chosen as a round number similar to our existing 256 matrix job limit, but we can very likely increase this as our benchmark + attempt counts grow.

**Note on Sandbox retries**

Sandboxes creation is subject to [dynamic quotas](https://vercel.com/docs/limits#dynamic-quotas) on vCPU allocation limits. The Sandbox SDK already implements a [basic retry policy](https://github.com/vercel/sandbox/blob/bf2bc66003fc89cf07a1346a7ea63951747cbec6/packages/vercel-sandbox/src/api-client/with-retry.ts#L55-L71), but it's not the most forgiving, so we add our own exponential backoff retries on top.

**Vercel project/creds**

Repo is currently wired up to the [evals-runner](https://vercel.com/supabase/evals-runner) project on Vercel via `VERCEL_PROJECT_ID`, `VERCEL_TOKEN`, `VERCEL_TEAM_ID`, a carryover from the Vercel Workflows spike. This project is configured w/ the 3 provider API keys.

If desired, we could move sandbox usage back to the evals project since we're not deploying any new application entrypoints with this PR.

**Job artifacts**

Previously each matrix job in `run-evals` uploaded a separate results artifact which we'd then download and combine. Now, since there's only one `run-evals` job we upload a single combined `raw-results` artifact.

CleanShot 2026-08-13 at 20 44 50@2x

**Job logs**

I intentionally don't wait for `command.logs()` because it sometimes loses connection and drops output. Instead, I consume `command.output('both')` so you'll see the `RUN` and `PASS`/`FAIL'` appear only at the end of each sandbox's eval execution, which seems like a reasonable tradeoff.

**Sample runs:**

- https://github.com/supabase/evals/actions/runs/31542233188
- https://github.com/supabase/evals/actions/runs/31543310069 (tests concurrency queuing)
- https://github.com/supabase/evals/actions/runs/31732096448 (codex gpt 5.4 mini, benchmark suite)
- https://github.com/supabase/evals/actions/runs/31735507954 (full benchmark refresh)

Results look normal after benchmark refresh ([Preview](https://evals-9wrtapw4g-supabase.vercel.app/)):

CleanShot 2026-08-13 at 17 02 22@2x

**How to test**

I test by manually dispatching the [refresh workflow](https://github.com/supabase/evals/actions/workflows/eval-refresh.yml) against this branch, though it's a slow / compute heavy thing to refresh the full suite, so you can probably just review the existing runs or dispatch a smaller refresh.

**Next steps**

Currently, benchmark execution takes ~33 min, comparable to [our latest refresh](https://github.com/supabase/evals/actions/runs/31185759729) on the GitHub Actions matrix. From here, we're planning to parallelize attempts instead of giving agents 2 attempts in sequence. Depending on how many sandboxes this ends up spawning, it'll potentially decrease execution time as attempts can run in parallel.

Another optimization I haven't landed here is the warm boot trick from Pedro's https://github.com/supabase/evals/pull/114, I'll likely revisit that as a follow up since it hasn't regressed overall benchmark duration yet.

**Related:**

- https://github.com/supabase/evals/pull/114
- https://github.com/supabase/evals/pull/186

Closes AI-912


apps/framework/scripts/run-vercel-evals.ts
```diff
@@ -0,0 +1,583 @@
...
+    },
+  });
+}
+
+/**
+ * Matches the SDK's retry policy of network errors, 429s, and 5xx.
+ * Other 4xx responses (bad token, invalid options) can never succeed.
+ * https://github.com/vercel/sandbox/blob/bf2bc66003fc89cf07a1346a7ea63951747cbec6/packages/vercel-sandbox/src/api-client/with-retry.ts#L10-L17
+ */
+export function isRetryableSandboxCreateError(error: unknown): boolean {
```
Rodriguespn: `isRetryableSandboxCreateError` returns `true` for every non-`APIError` ([`:340`](https://github.com/supabase/evals/pull/192/changes#diff-8e35b802b923520e05500a9549c95311442f6b6d37ec79db609713fec57f10f1R340)), so a deterministic client-side failure, like bad token, invalid options, `Could not get credentials from OIDC context`, is treated as retryable and spun 12× on backoff. 

Because the pair-level `pRetry(retries: 2)` ([`:108`](https://github.com/supabase/evals/pull/192/changes#diff-8e35b802b923520e05500a9549c95311442f6b6d37ec79db609713fec57f10f1R108)) wraps `createSandbox`'s `pRetry(retries: 12)` ([`:313`](https://github.com/supabase/evals/pull/192/changes#diff-8e35b802b923520e05500a9549c95311442f6b6d37ec79db609713fec57f10f1R314)), the two layers **multiply**: 

I let Claude reproduce this by spinning 39 `Sandbox.create` attempts over 588s (~9.8 min)** for one unrecoverable credential error. At `concurrency: 250` a single bad CI secret makes all 166 pairs thrash ~10 min before the job fails with a wall of identical lines. The comment claims parity with the SDK, but the SDK only retries network errors + 429 + 5xx ([with-retry.ts](https://github.com/vercel/sandbox/blob/main/packages/vercel-sandbox/src/api-client/with-retry.ts)); 

Suggest: in `isRetryableSandboxCreateError`, retry a non-`APIError` only when it's a genuine network error (undici `fetch failed` / `ECONNRESET`-style cause), not any `Error`; and give the outer pair `pRetry` a `shouldRetry` so it doesn't re-wrap failures the inner layer already declared terminal (see [p-retry `shouldRetry`](https://github.com/sindresorhus/p-retry#shouldretry)).

apps/framework/scripts/run-vercel-evals.ts
```diff
@@ -0,0 +1,583 @@
...
+  return value;
+}
+
+type VercelCredentials = ReturnType<typeof vercelCredentialsFromEnv>;
+
+/** Returns explicit credentials because the SDK does not infer all local vars. */
+function vercelCredentialsFromEnv():
+  | { token: string; teamId: string; projectId: string }
+  | Record<string, never> {
+  if (process.env.VERCEL_OIDC_TOKEN) return {};
```
Rodriguespn: `if (process.env.VERCEL_OIDC_TOKEN) return {}` defers to OIDC whenever any Vercel OIDC token is present, even when explicit `VERCEL_TOKEN`/`VERCEL_TEAM_ID`/`VERCEL_PROJECT_ID` are set. 

Having a stale `VERCEL_OIDC_TOKEN` locally is easy (left by `vercel env pull` or the Vercel CLI plugin); it then shadows the real token and every create fails with `Could not get credentials from OIDC context`. 

I would prefer explicit `VERCEL_TOKEN` when it's non-empty and fall back to OIDC only when it's absent. 

PS: The `{ token; teamId; projectId } | Record<string, never>` return type is an awkward "or nothing". A small `?` optional/discriminated shape reads clearer.

.github/workflows/eval-refresh.yml
Rodriguespn, line 260: nit: Now that the matrix collapsed to one long-lived controller job: 

(a) Suggest to add an explicit `timeout-minutes:` (for example ~90). There's none, so it defaults to GitHub's 360-min ceiling and a wedged controller could hold a runner for 6h. 

(b) The checkout uses `submodules: recursive` ([`:275`](https://github.com/supabase/evals/pull/192/changes#diff-943b3ca0618de4aae9edbfa6d5f40c7409050c29c6802f3a9752ebde3d76a03bL267)), but the controller never touches submodules. The sandbox clones and inits its own inside the VM. Seems to be avoidable checkout work.

apps/framework/scripts/run-vercel-evals.ts
```diff
@@ -0,0 +1,583 @@
...
+        'url.https://github.com/.insteadOf=git@github.com:',
+        'submodule',
+        'update',
+        '--init',
+        '--recursive',
+      ],
+      cwd: sandbox.cwd,
+      timeoutMs: 2 * 60 * 1_000,
+    });
+    await runSandboxCommand(sandbox, label, 'install Docker', {
```
Rodriguespn, line 203: Possible follow up: Every pair/attempt re-does cold setup (Docker setup, install pnpm, install dependencies, etc...). 

Two suggestions, both deferred by the spike but worth tracking: 

1) a warm-boot snapshot pre-baking docker + `node_modules` + Supabase images (example from my PR[#114](https://github.com/supabase/evals/pull/114)); I'm currently implementing this on a follow up PR to check the actual net gain from snapshotting the sandbox images.

2) parallel attempts, and pass if one of the runs succeeds over runs or the amount of successful runs is higher than a predefined threshold (2 out of 3 pass, for example), instead of `run-eval`'s sequential stop-on-pass retry. Worth considering as we [discussed offline](https://supabase.slack.com/archives/C051L8U2EJF/p1786654709621309?thread_ts=1786399252.414279&cid=C051L8U2EJF)

Rodriguespn:
nit: We should clean the `run-evals-sandboxes` and the other related Github labels that I added while working on https://github.com/supabase/evals/pull/192

apps/framework/scripts/run-vercel-evals.ts
mattrossman: I've updated this in https://github.com/supabase/evals/pull/192/commits/5b11a9b8545213d538307ddcaf03262053a61f11 and https://github.com/supabase/evals/pull/192/commits/8fa6fb39c9ecce8ed671ee2adcbffecfb8c2e063 with 3 parts:

- Known OIDC credential errors now bail immediately
  -  This also bails the outer layer now by leveraging `p-retry`'s `AbortError`
- Known network errors follow the normal 12 retry limit as `APIError`
- Unrecognized errors have a smaller, 2 retry limit, and log a warning so we can consider adding cases for them

The reasoning for the large 12 retry limit is Sandbox SDK's dynamic quotas, that ramp up vCPU allocation limits as your usage increases, so it's important we're resilient to flaky sandbox creation while ramping up.

There might be other transient errors besides the ones I've accounted for, or other always-failing errors, hance the compromise on a 2-retry allowance on unknown errors.

apps/framework/scripts/run-vercel-evals.ts
mattrossman: Looking at this again, I don't think I need `VERCEL_OIDC_TOKEN` support here in the first place, I think that was carryover from an earlier PR.

- This is only helpful if devs do `vercel link` and `vercel env pull`, which feels awkward since you'd be linking to the secondary "evals-runner" project in a workspace otherwise associated with the "evals" project, and we don't document this workflow (I've never tested it either)
- Those local dev OIDC tokens expire after 12h
- This code path is for create sandboxes, but in local dev you'd normally run evals locally on your machine

Removing the OIDC auth path also simplifies the return type, and would trim some of the work I did [above](https://github.com/supabase/evals/pull/192#discussion_r3784329211) on sandbox creation retry handling.

Updated in https://github.com/supabase/evals/pull/192/commits/6d74fa1dd90e14e7caff58fd9912949aff41fa12

.github/workflows/eval-refresh.yml
mattrossman: I've added a limit of 60 minutes (shorter than your suggestion, b/c I'd like to target evals finishing well below an hour) and removed the leftover submodule init part from our earlier GH-based runner.

I'm also modifying the workflow to make timeouts easier to detect and recover from, see [sample run](https://github.com/supabase/evals/actions/runs/31814494200).

- `publish-results` now runs even if `run-evals` fails or times out, so completed pairs still get committed to results snapshot
- An "Audit pair results" step warns if a run was incomplete, both as a job annotation and in the job summary
- "Merge scheduled results pull request" skips when the audit finds gaps, so nightly regression refreshes don't auto-merge an incomplete run
- A "Require complete results" step turns the run red at the end if results were incomplete, so it can't be mistaken for a clean run

CleanShot 2026-08-14 at 11 44 14@2x

CleanShot 2026-08-14 at 11 44 22@2x

mattrossman:
> We should clean the run-evals-sandboxes and the other related Github labels

@Rodriguespn I've removed the 2 sandbox related labels, along with the stale AI gateway opt-in label

apps/framework/scripts/run-vercel-evals.ts
mattrossman: Agreed on exploring both, resolving this for now as we can handle those in follow up PRs

Rodriguespn (approved):
LGTM, thanks for addressing the feedback 🙌 Great work adding a partial results recovery mechanism!

PR #193, opened 2026-08-13
feat(core): launch a local mcp build via SUPABASE_MCP_SERVER_PATH

barryroodt:
Stack 1/4 for #128. Base: `main`. (`compare` moved out of the chain into #195, parked on #194.)

Re-lands the core of #118 without the submodule, so a workspace can test an unpublished [`mcp-server-supabase`](https://github.com/supabase/mcp) change against the evals without publishing to npm. Carries the review fixes from that thread (canonicalize the base once, receipt-visible mounts) and the tests. Thanks @Rodriguespn.

Set `SUPABASE_MCP_SERVER_PATH` and `createConfig` launches that build with `node` instead of the pinned package with `npx`. It takes a repo dir, a package dir, or a direct `.js`/`.mjs`/`.cjs` entrypoint, and resolves relative paths against the evals checkout root rather than the process CWD.

## Three details worth your attention

These are what the tests pin, and each exists because a CLI agent spawns its MCP server **inside** the sandbox container, which inherits nothing from the harness process.

- **`node`, not `process.execPath`.** The host's absolute node path does not exist in-container, so both sides resolve `node` through `PATH`.
- **The base is canonicalized once, and the entry derived from it.** Docker resolves bind-mount sources against the daemon's view, so command and mount must agree on one filesystem view. Realpathing the entry separately would let a symlinked `dist/` resolve outside the mounted directory.
- **The mount covers the build's whole git toplevel**, not just `dist/`, because the build is unbundled and needs its `node_modules` at runtime.

Hence `SandboxMount` and `supabaseMcpServerMounts()`, threaded `local-stack-runtime` -> `agent-environment` -> `docker-sandbox` as `--volume host:host:ro`. Mounting at the identical real path is what lets one command config work on the host and in the container.

## Nothing changes when the var is unset

`supabaseMcpServerMounts()` returns `[]`, and I checked the config rather than assuming: ran `createConfig({apiUrl})` against `main` and against this branch with the var cleared, and diffed. Identical, matching md5 `875658e9…`, same `npx` command and same seven args in the same order.

## How to verify

```bash
git submodule update --init && pnpm install

# zero cost
pnpm --filter @supabase-evals/core test      # 114/114
pnpm --filter @supabase-evals/sandbox test   # 39 passed, 4 skipped
pnpm typecheck && pnpm format:check

# end to end (costs one agent run): make a visible change to a tool response in
# your mcp checkout, `pnpm install && pnpm build` there, then
SUPABASE_MCP_SERVER_PATH=<mcp-checkout> pnpm eval --eval investigate-db-001-table-row-counts --experiment claude-haiku-4.5
```

Unset the var and the config is byte-identical to `main`, which is what the new core tests pin.

## Verified

- core 114/114 (10 new), sandbox 39 passed, 4 skipped (counts after the rebase onto current `main`)
- `tsc --noEmit` and `biome check` clean

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,11 +1,17 @@
 import vm from 'node:vm';
 import { createRequire } from 'node:module';
 import { createHash, createHmac } from 'node:crypto';
-import { execFile } from 'node:child_process';
+import { execFile, execFileSync } from 'node:child_process';
 import { createServer } from 'node:net';
 import { promisify } from 'node:util';
-import { existsSync, mkdtempSync, readFileSync, rmSync } from 'node:fs';
-import { basename, dirname, join } from 'node:path';
+import {
+  existsSync,
+  mkdtempSync,
+  readFileSync,
+  realpathSync,
+  rmSync,
+} from 'node:fs';
+import { basename, dirname, join, resolve } from 'node:path';
 import { tmpdir } from 'node:os';
 import { fileURLToPath } from 'node:url';
 import type { ToolCall, ToolName } from './transcript/types.js';
@@ -409,6 +415,20 @@ export type AgentHarness = {
  */
 export type SkillSource = { name: string; dir: string };
 
+/**
+ * A host directory bind-mounted into the agent sandbox. Read-only by default;
+ * mounted at the identical container path unless `containerPath` overrides it
+ * (identical paths let one command config work on both host and container).
+ */
+export type SandboxMount = {
+  /** Host directory to mount. */
+  hostPath: string;
+  /** Mount point inside the container; defaults to `hostPath`. */
+  containerPath?: string;
+  /** Mount read-only (default true). */
+  readonly?: boolean;
+};
+
 export type LocalStackSessionArgs = {
   /** Supabase CLI version this scenario requires, overriding the runtime default. */
   cliVersion?: string;
@@ -444,6 +464,12 @@ export type LocalStackSessionArgs = {
    * instead, so they ignore this.
    */
   skills?: readonly SkillSource[];
+  /**
+   * Extra host directories to bind-mount into the sandbox (read-only by
+   * default) — e.g. a local MCP server build the in-container agent must be
+   * able to launch. See `supabaseMcpServerMounts`.
+   */
+  mounts?: readonly SandboxMount[];
   /**
    * Skip installing the real Supabase CLI into the sandbox (from
    * `skipCliInstall:` frontmatter), for scenarios whose prompt has the agent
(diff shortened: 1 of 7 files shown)


packages/core/src/index.ts
```diff
@@ -924,12 +951,111 @@ export function supabaseMcpServer(
...
+  const realBase = realpathSync(base);
+  const baseDir = isEntryFile ? dirname(realBase) : realBase;
+  const value: LocalMcpServer = {
+    entry: isEntryFile
+      ? realBase
+      : join(realBase, 'dist', 'transports', 'stdio.js'),
+    baseDir,
+    // The whole git toplevel (not just dist/) because the build is unbundled:
+    // it requires its node_modules at runtime.
+    mountRoot: gitToplevel(baseDir) ?? baseDir,
```
mattrossman, line 1046: Wondering if we should make this more constrained to only the files necessary? I think just `node_modules/`, `dist/`, and `package.json`.

Concern is the local checkout can contain secrets in `.env` and IDK if we want agents poking through that / exposing it in transcripts. Granted even w/ that tightened they can print LLM keys from environment variables currently, but still seems like a low hanging fruit to reduce risk.

mattrossman:
Tested by patching `search_docs` to return a canary string, and confirmed it came through in tools mode and sandbox mode:
- `investigate-db-001-table-row-counts` w/ claude-haiku-4.5
- `build-cli-002-declarative-schema` w/ `claude-code-sonnet-5`

mattrossman (approved):

packages/core/src/index.ts
barryroodt: Fair concern, and aimed at the right thing. The specific narrowing won't hold though: `packages/mcp-server-supabase/node_modules/zod` and `@supabase/mcp-utils` are pnpm symlinks pointing outside the package, into the root `.pnpm` store and into a sibling workspace package. Mount only that package's `node_modules/`, `dist/` and `package.json` and both dangle in the container.

A narrowed mount that works needs the repo root `node_modules` and `packages/mcp-utils` on top of the server's own three, and that list has to keep pace with the dependency graph.

I checked my `supabase/mcp` checkout for `.env` files (`git status --porcelain --ignored`) and there are none. One checkout, so not a guarantee for everyone's tree.

My call is to keep the single read-only toplevel mount. If you want the hardening anyway, we can mask specific paths with empty read-only binds over the toplevel mount, which costs the same path list without breaking resolution. I'd switch if a real `.env` turns up in the repo's normal layout, or if the sandbox grows a way to exclude subpaths without a maintained list.

mattrossman: Works for me, FWIW my local clone has secrets in `packages/mcp-server-supabase/.env.local` used for the E2E tests but risk is pretty minimal here so I'm okay keeping it as-is.

PR #194, opened 2026-08-13
feat(eval): add strict local-run gates and provenance receipts

barryroodt:
Stack 2/4 for [#128](https://github.com/supabase/evals/issues/128). Base: [#193](https://github.com/supabase/evals/pull/193).

This keeps `pnpm eval` as the single entrypoint for local and CI runs:

```bash
pnpm eval -- <eval...> --experiment <id> --strict [--mcp <path>]
```

- `--strict` turns error-class skips into exit 1, and CI enables it.
- Existing runner paths own experiment validation and per-agent readiness. Judge keys and declared `SKILL.md` files fail before spend.
- `--mcp` uses the override mechanism from #193.
- Canonical results in `results/<experiment>/<eval>.json` include provenance and use atomic replacement. Corrupt partial results rerun.

The separate `pnpm local` CLI, `results-local/`, and published-result helpers are gone. `pnpm eval -- list` remains the discovery command.

Verification: smoke 14/14, core 114/114, sandbox 39/39, typecheck + format passed.

The pre-existing `test:framework` suite reached 8 PASS markers, then stopped at `investigate-security-001-public-table` because `OPENAI_API_KEY` was absent. A credentialed run still needs:

```bash
pnpm --filter @supabase-evals/framework test:framework
```


apps/framework/scripts/local.ts
```diff
@@ -0,0 +1,458 @@
...
+    '-1',
+    PUBLISHED_LOG_FORMAT,
+    '--',
+    file,
+  ]);
+  return { file, rows, ...parsePublishedLog(line) };
+}
+
+/** Fetch origin/main so the published exports are current; warn-and-continue offline. */
+function fetchMain() {
```
mattrossman: Is this called anywhere?

apps/framework/scripts/local.ts
```diff
@@ -0,0 +1,458 @@
...
+/**
+ * The agent itself needs its provider key. Checked here (not just by the
+ * harness) because the harness SKIPs the experiment with exit 0 on missing
+ * credentials — the runner would only notice at the no-result check. A
+ * set-but-EMPTY var counts as missing (node --env-file does not override
+ * an existing env var, even an empty one, so a stray `export KEY=` in the
+ * shell silently shadows .env — observed live).
+ */
+function validateAgentKey() {
+  if (process.env.ANTHROPIC_API_KEY) return;
```
mattrossman: Is the intent to check all LLM provider keys that agents might use? Currently this checks only the Anthropic key.

Note we already [check for API keys](https://github.com/supabase/evals/blob/a2325afd0617e0e83c0c35c5dbfa22bca59417e4/packages/core/src/agents/engine.ts#L77-L79) with per-agent variable names on `runner.apiKeyEnvVar`, this looks like it overlaps with that.

apps/framework/scripts/local.ts
```diff
@@ -0,0 +1,458 @@
...
+}
+
+/**
+ * Evals whose scorer uses the LLM judge grade the agent's output with an
+ * OpenAI model — even when the agent under test is Claude. A missing grader
+ * key otherwise surfaces only AFTER the (paid) agent run, wasting it.
+ * Textual scan of EVAL.ts; a false positive just asks for a key early.
+ */
+function validateJudgeKeys(evalIds: string[]) {
+  if (process.env.OPENAI_API_KEY) return;
```
mattrossman: This seems like a good thing to check. It'd be nice to fix this lower down so we're not wasting tokens even if running via `pnpm eval` without the `pnpm local` wrapper.

apps/framework/scripts/local.ts
```diff
@@ -0,0 +1,458 @@
...
+}
+
+// ---------- subcommands ----------
+
+async function cmdExperiments() {
+  const published = new Set(
+    loadPublished().flatMap((f) => f.rows.map((r) => r.experiment))
+  );
+  console.log(
+    `${'EXPERIMENT'.padEnd(36)} ${'AGENT'.padEnd(12)} ${'MODEL'.padEnd(22)} ${'EFFORT'.padEnd(8)} PUBLISHED`
```
mattrossman: What's the motivation for this `PUBLISHED` column? It seems like this is mainly related to the `pnpm local compare` subcommand proposed in https://github.com/supabase/evals/pull/195

Wondering if this column and its helpers should be pruned here.

apps/framework/scripts/local.ts
```diff
@@ -0,0 +1,458 @@
...
+  case 'run':
+    await cmdRun(rest);
+    break;
+  case 'experiments':
+    await cmdExperiments();
+    break;
+  default:
+    fail(`usage: pnpm local <run|experiments> ...
+  run <eval...>   run eval(s) in your world (skills tree as-is; --mcp override)
+  experiments     list experiments (agent, model, effort, published availability)`);
```
mattrossman: Could you help me understand the motivation behind the new `pnpm local` entrypoint with these subcommands? 

Today we have the `pnpm eval` entrypoint, so I'm wondering why we're not putting these gates and local repo orchestration in there.

Also trying to understand the motivation for `pnpm local experiments`, when would I invoke that versus looking inside of `experiments/`? The output partially overlaps with the `pnpm eval -- list` command, used in CI for experiment discovery, so if the goal is a human readable version I wonder if they can be merged.

apps/framework/scripts/local.ts
```diff
@@ -0,0 +1,458 @@
...
+/** Load every published export once; callers share the result. */
+function loadPublished(): PublishedFile[] {
+  return Object.values(PUBLISHED_EXPORTS).flatMap(
+    (f) => loadPublishedFile(f) ?? []
+  );
+}
+
+// ---------- eval validation (fail before spending) ----------
+
+function validateEvals(evalIds: string[]) {
```
mattrossman: Seems like this overlaps with existing validation in `run-eval`, which fails on invalid eval IDs ([1](https://github.com/supabase/evals/blob/a2325afd0617e0e83c0c35c5dbfa22bca59417e4/apps/framework/harness/run-eval.ts#L685-L692)) or parsing issues with prompt frontmatter ([2](https://github.com/supabase/evals/blob/a2325afd0617e0e83c0c35c5dbfa22bca59417e4/apps/framework/harness/run-eval.ts#L128-L131)) during discovery

apps/framework/scripts/local.ts
```diff
@@ -0,0 +1,458 @@
...
+      );
+    } catch (err) {
+      fail(
+        `eval metadata invalid — fix evals/${id}/PROMPT.md before spending on runs\n${err instanceof Error ? err.message : String(err)}`
+      );
+    }
+  }
+}
+
+function validateExperiment(experiment: string) {
```
mattrossman: Similar to above [comment](https://github.com/supabase/evals/pull/194/changes#r3798696443), overlaps with existing experiment discovery/validation ([here](https://github.com/supabase/evals/blob/a2325afd0617e0e83c0c35c5dbfa22bca59417e4/apps/framework/harness/run-eval.ts#L656-L665))

apps/framework/scripts/local.ts
```diff
@@ -0,0 +1,458 @@
...
+  }
+  return p;
+}
+
+/**
+ * The experiment's declared skills must exist in this checkout, or the
+ * treatment silently runs skill-less against a skills-enabled published
+ * baseline — a world mismatch, not a comparison.
+ */
+async function validateSkills(experiment: string) {
```
mattrossman: Good idea to validate this! 👍 Worth consolidating w/ our existing [warning](https://github.com/supabase/evals/blob/a2325afd0617e0e83c0c35c5dbfa22bca59417e4/apps/framework/harness/run-eval.ts#L166-L171) for missing skills.

.gitignore
```diff
@@ -12,3 +12,5 @@ evals/*/local/supabase/.branches/
 results/*/
 .sync-tmp/
 
+# local-dev runner (apps/framework/scripts/local.ts)
+/results-local/
```
mattrossman: Do we need a separate folder for this, since (A) the `results/` folder is gitignored already and (B) we're adding provenance receipts so you can distinguish locally generated results?

mattrossman:
I think some of the complexity here comes from legacy behavior where there's an opportunity to step back and simplify.

At one point (when evals primarily ran in a shared process) it was desirable for evals to fail passively, so that it wouldn't terminate the whole workflow in CI. That's why a missing credential still shows SKIP + exit 0 instead of failing the run.

Since then, we've moved to a GitHub Actions matrix with parallel processes, and further still to Vercel Sandboxes. In both cases, we only run one eval pair per runner/sandbox. Now we actually prefer behavior where bad evals fail loudly so we know which pairs need to be re-run.

I think this translates into a couple simplification opportunities:

- Keep `pnpm eval` the entrypoint, and put the gates there instead of a second entrypoint
- Add a `--strict` flag to `pnpm eval` that turns failures into `exit 1` instead of passive `SKIP`, and enable this in CI

That along w/ the other notes above could shave a couple hundred LOC off the PR.

.gitignore
barryroodt: Good catch. Removed the separate folder.

apps/framework/scripts/local.ts
barryroodt: No, it was only for the compare path. Removed it.

apps/framework/scripts/local.ts
barryroodt: Good catch, removed.

`apps/framework/harness/run-eval.ts` now owns eval discovery and prompt parsing.

apps/framework/scripts/local.ts
barryroodt: ok great, thanks! Fixed.

apps/framework/scripts/local.ts
barryroodt: Agreed, fixed.

apps/framework/scripts/local.ts
barryroodt: Done. The existing warning remains for normal runs, while strict preflight now requires each declared `skills/<name>/SKILL.md` before spend

apps/framework/scripts/local.ts
barryroodt: Agreed, that column belonged to compare. Removed it.

apps/framework/scripts/local.ts
barryroodt: Original intent was to keep it explicitly separate, but in hindsight it may have been overkill.

Fixed so that `pnpm eval` is the single entrypoint now.  
`pnpm eval -- list` remains the discovery command.

barryroodt:
> I think some of the complexity here comes from legacy behavior where there's an opportunity to step back and simplify.
> 
> At one point (when evals primarily ran in a shared process) it was desirable for evals to fail passively, so that it wouldn't terminate the whole workflow in CI. That's why a missing credential still shows SKIP + exit 0 instead of failing the run.
> 
> Since then, we've moved to a GitHub Actions matrix with parallel processes, and further still to Vercel Sandboxes. In both cases, we only run one eval pair per runner/sandbox. Now we actually prefer behavior where bad evals fail loudly so we know which pairs need to be re-run.
> 
> I think this translates into a couple simplification opportunities:
> 
> * Keep `pnpm eval` the entrypoint, and put the gates there instead of a second entrypoint
> * Add a `--strict` flag to `pnpm eval` that turns failures into `exit 1` instead of passive `SKIP`, and enable this in CI
> 
> That along w/ the other notes above could shave a couple hundred LOC off the PR.

Thanks for the feedback. 

I've now implemented this shape. 
- `pnpm eval -- --strict` turns error-class skips into exit 1, and the refresh workflow enables it in [`4604995`](https://github.com/supabase/evals/commit/4604995). 
- moved the judge-key check into `run-eval`, so plain runs now fail before spend when a required judge key is missing ([`c100322`](https://github.com/supabase/evals/commit/c100322)). 
- unknown CLI arguments now fail loudly instead of being silently ignored ([`a61e6f1`](https://github.com/supabase/evals/commit/a61e6f1)).

One boundary: a published pair that does not exist in the current checkout stays a visible `SKIP`, including under strict mode. That lets an older branch run after `main` publishes a newer eval; malformed pairs present in the checkout still fail.

PR #205, opened 2026-08-14
docs: cover example solutions and check design for scorers

czenko:
Closes DOCS-1324

## Problem

Ten more docs evals are scoped. The RLS eval took several review rounds, and most of the findings were scorer defects a contributor can catch before asking for review. `CONTRIBUTING.md` doesn't say how. Guidance comes from [the review of #168](https://github.com/supabase/evals/pull/168#pullrequestreview-4938302559).

## Solution

- Adds **Checking your scorer**: one green solution that passes every check, a few bad ones carrying one deliberate flaw each, and the expected failures written down before scoring. Split into solution shapes, what to write, how to score, and reading the results.
- Names three solution shapes, since the 44 evals don't share one. A workspace overlay for local-stack evals, a `remote/`-shaped seed for tools evals, and a report for the 8 evals scored on what the agent said rather than what it built.
- Adds **Designing checks**: decide what a check does when the object is absent, and find a class of object by asking the system that owns it. Checks about an object the prompt names can name it.
- Notes that a flaw usually trips several checks, and that transcript checks can't be exercised by a solution you wrote yourself.
- Lists `solutions/` in the eval folder contents, above the optional entries, and links the new section from **Submitting evals for review**.

## Manual testing

1. Open [CONTRIBUTING.md on this branch](https://github.com/supabase/evals/blob/mirandalimonczenko/docs-1324-document-example-solutions-and-check-design-in-the-evals/CONTRIBUTING.md). **Designing checks** and **Checking your scorer** sit between **Writing scorers** and **Adding an experiment**.
2. Click either `Checking your scorer` link in that page. Both jump to the new section.
3. Read the diff of the new content. See that it is technically accurate and helpful for all Eval contexts.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,8 +13,9 @@ Then add a folder under `evals/` containing:
 
 1. `PROMPT.md` with frontmatter metadata and the task the agent sees.
 2. `EVAL.ts` with the scorer.
-3. Optional `remote/` data when the scenario needs to seed hosted project state, such as database, logs, or functions.
-4. Optional `local/` files when the scenario needs to seed a local filesystem, such as a local `supabase/` project.
+3. `solutions/` with the example solutions you scored the eval against. See [Checking your scorer](#checking-your-scorer).
+4. Optional `remote/` data when the scenario needs to seed hosted project state, such as database, logs, or functions.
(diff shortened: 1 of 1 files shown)


czenko:
Closing based on what @mattrossman said:
> it's mostly the solutions/ part I'm hesitant to adopt rn as it seems like it will increase volume of stuff to review per scenario.

PR #206, opened 2026-08-14
feat(framework): score committed solutions without an agent run

czenko:
Closes AI-1080

Stacked on #205. Review that one first, or read this diff against its branch.

## Problem

Checking whether a scorer's checks behave takes an agent run. That run is nondeterministic and costs a model call, so you can't separate a scorer defect from agent variance, and a check that passes broken work is hard to catch before the eval ships. #205 tells contributors to score example solutions. Nothing scored them.

## Solution

- Adds `pnpm eval:solution -- --eval <id>`, which boots the eval's environment, copies a directory from its `solutions/` over the workspace, and runs the scorer with no agent. `--solution <name>` scores one.
- Copies the overlay between `localDir` and `supabase start`, so a solution's migrations apply on boot with no per-eval apply step. That is the whole mechanism.
- Moves eval discovery and run setup to `eval-discovery.ts`, so the agent and solution entry points can't drift on what an eval is.
- Writes nothing to `results/`. Exported workspaces land in `.solution-runs/`, gitignored.
- Ships two solutions for `build-cli-002-declarative-schema` and the expected-failure list that goes with them.
- Documents the command in `README.md` and in the guidance from #205.

Local-stack evals only. A tools eval's starting state is a seeded hosted project, not a workspace to copy files into, so it needs its own apply path.

## Manual testing

1. Stop any local Supabase stack, since the run needs ports 54321-54329.
2. Score both solutions:

   ```bash
   pnpm eval:solution -- --eval build-cli-002-declarative-schema
   ```

   Output matches `solutions/README.md`. `green` fails only the tool-call check, `migration-only` also fails the schema-file check:

   ```
   green
     ❌ supabase db diff used to generate the migration
     ✅ schema file updated to include description column
     ✅ a new migration was generated for the change
     ✅ description column exists in the live database
     3/4 checks passed

   migration-only
     ❌ supabase db diff used to generate the migration
     ❌ schema file updated to include description column
     ✅ a new migration was generated for the change
     ✅ description column exists in the live database
     2/4 checks passed
   ```

3. Run `git status results/`. It is clean, and `.solution-runs/build-cli-002-declarative-schema/` holds both exported workspaces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,5 @@ evals/*/local/supabase/.branches/
 .DS_Store
 results/*/
 .sync-tmp/
+.solution-runs/
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Then add a folder under `evals/` containing:
 
 1. `PROMPT.md` with frontmatter metadata and the task the agent sees.
 2. `EVAL.ts` with the scorer.
-3. `solutions/` with the example solutions you scored the eval against. See [Checking your scorer](#checking-your-scorer).
+3. `solutions/` with the example solutions you scored the eval against, one directory per solution. See [Checking your scorer](#checking-your-scorer).
 4. Optional `remote/` data when the scenario needs to seed hosted project state, such as database, logs, or functions.
 5. Optional `local/` files when the scenario needs to seed a local filesystem, such as a local `supabase/` project.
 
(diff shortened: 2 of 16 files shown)


czenko:
Perhaps this is beyond my scope to recommend. Closing this for now. Pinning it as a thought. I may run this locally to evaluate my own work.

PR #212, opened 2026-08-17
feat: eval the API keys guide in supabase.com/docs

czenko:
Closes DOCS-1301

## Problem

We point agents at the API keys guide. Nobody has measured whether reading it is enough to keep the secret key out of a browser bundle.

No existing eval answers that. `resolve-sdk-001-legacy-key-migration` is Node scripts, server-side throughout. `build-functions-004/005/006` are Edge Function auth, where the secret key is legitimately server-side. None of them builds a client bundle, so none can catch a secret key shipped to a browser.

## Solution

- Adds `build-docs-003-api-keys-guide`. The guide is the subject under test. The prompt is a product request plus the guide's url, and never says key, publishable, secret, service role, server, expose, or environment variable.
- Seeds a Vite app with two screens. Sign-up is public and correct with a publishable key. The roster lists everyone who signed up with their email, which lives in `auth.users` and no publishable key reaches. The roster is the one place where the convenient answer ships the whole table to the browser.
- Scores nine checks. Six read the built bundle and client source, two probe the running stack, one confirms the guide was retrieved.
- Adds `ctx.stackStatus()` to the local-stack scoring context, which reports the stack's url plus its publishable, secret, and legacy anon keys. The scorer needs all four, and `getClient()` only carries the publishable key.

### Checks that exist to stop a false green

- **`roster returns every signed-up email`** pairs with the bundle checks, which a run that built nothing passes trivially.
- **`client source contains a signUp call`** pairs with `client bundle carries a publishable or anon key`, which passes for a key that ships and is never called.
- **`no client role can read another user's email`** catches the design that dodges the elevated call entirely by copying addresses into a table anyone can read. Everything else scored green before this check existed.

### Where the email check came from

Not from the page under test. The page frames publishable-key safety as an RLS checklist, and the shortcut of copying addresses into a readable table satisfies that checklist while leaking. It came from `api/securing-your-api`, which treats grants and RLS as separate layers.

### What this eval does not score

Two checks were tried and removed, because both measured the platform rather than the page. Keeping either means a CLI upgrade reports as a docs change.

- **Which key format the client uses.** The build injects a placeholder over `VITE_SUPABASE_ANON_KEY`, so a client wired through that name never carries a real key into the bundle and the format is not observable.
- **Which key format the server uses.** The Edge Function runtime decides which keys a function is handed, and the pinned CLI hands over legacy ones only. A function reading them follows the runtime, not the guide.

## Known limitations

- **The roster probe seeds two users and requires both addresses in one response.** An agent that paginates to fewer than two rows per page would fail on a design choice, not a defect. `auth.admin.listUsers()` defaults to 50 per page and PostgREST to 1000 rows, so this is a narrow edge, but it would surface as an unexplained red.
- **`no secret-bearing env var is client-exposed` passes by design.** It reads every `.env` outside `supabase/` and fails on a secret in any of them. An agent that keeps the secret in the function env passes, and that is the intended outcome.
- **`client source contains a signUp call` is a literal match.** Dead code satisfies it, and a call reached only from outside client source does not. Proving the screen works needs a driven DOM the scorer does not have.
- **Authorization is out of scope.** The roster probe calls as a signed-in user, so a roster gated on sign-in counts as working. Restricting the endpoint to staff needs a role in the seed and is what `build-rls-003-org-roles-permissions` measures.
- **`suite: regression`,** not benchmark. It is a regression detector for the guide, and it reports no failure against the published page.

## Baseline

`claude-code-sonnet-5` with and without skills, 3 runs each, [CI run 33549257417](https://github.com/supabase/evals/actions/runs/33549257417): **9/9 on all six runs.** Results are committed to this branch.

No check fails against the published guide, which is the floor a regression detector needs. A later docs change that breaks key placement shows up as a drop from this.

The scorer was also pointed at a revised draft of the guide, and that came back 9/9 on all six runs too, so the draft introduces no regression.

## Manual testing

1. Free the Supabase ports. Run `docker ps --format '{{.Names}}\t{{.Ports}}' | grep -E '5432[12]->'` and stop whatever holds them.
2. Run `pnpm eval:dry -- --eval build-docs-003-api-keys-guide`. It plans 18 experiments.
3. Run `pnpm typecheck`. Clean.
4. Read `README.md` in the eval folder. Confirm the persona rules and the check rationale match the scorer.
5. The scorer was checked against thirteen throw-away solutions with no agent in the loop. A correct one passes every check except the transcript check, which a no-agent run always fails, and each bad one fails the checks it was built to fail.


evals/build-docs-003-api-keys-guide/PROMPT.md
```diff
@@ -0,0 +1,28 @@
...
+  - data-api
+topic:
+  - sdk
+  - security
+services:
+  - gotrue
+  - kong
+  - postgrest
+  - edge-runtime
+skills: []
```
czenko:
```suggestion
```
Previous decision: we also run this with skills. Can't hurt.

mattrossman:
(Merging main in so that it can use the new 3-run averaging strategy)

evals/build-docs-003-api-keys-guide/access.ts
```diff
@@ -0,0 +1,173 @@
...
+): Promise<CheckResult> {
+  const name = 'roster returns every signed-up email';
+  let body: string;
+  let httpStatus: number;
+
+  try {
+    const res = await fetch(`${status.apiUrl}/functions/v1/${ROSTER}`, {
+      headers: {
+        apikey: status.publishableKey,
+        authorization: `Bearer ${status.publishableKey}`,
```
mattrossman: When testing locally, this check failed because the agent restricted roster access to signed-in users by gating on `auth.getUser()`. Since the prompt doesn't specify who can see the roster, I'd expect that to pass (e.g. by checking with `clientA`'s token), or the prompt to be more clear about it being publicly visible. I know you mention that full auth / role checking is out of scope, but in this case the agent didn't check for any particular user info, just that they're signed in.

mattrossman: I reproduced this in CI in the latest `run-evals-changed` run, see failing check [here](https://github.com/supabase/evals/blob/770669e2197cb026db0f065d94d8b3ed4f8e4861/apps/web/src/data/regression-eval-results.json#L347-L351) which mentions a "Not signed in" error in the notes.

evals/build-docs-003-api-keys-guide/EVAL.ts
```diff
@@ -0,0 +1,78 @@
...
+    const setup = await setupFixtures(ctx);
+    const access: CheckResult[] =
+      'failure' in setup
+        ? [setup.failure]
+        : [
+            await checkRoster(status, setup.fixtures),
+            await checkEmailsHiddenFromClients(ctx, setup.fixtures),
+          ];
+
+    const checks: CheckResult[] = [
```
mattrossman, line 22: Nit: I find it helpful to see the full list of checks declared here, splitting it into separate files and spreading obscures that a bit.

evals/build-docs-003-api-keys-guide/bundle.ts
```diff
@@ -0,0 +1,228 @@
...
+      name: 'client source calls signUp',
+      passed: signUpWired,
+      notes: signUpWired
+        ? undefined
+        : 'no auth.signUp call in client source, so nothing creates an account',
+    },
+    {
+      // Passes when there is no secret-bearing env var at all, which is one
+      // valid design rather than something the scenario requires.
+      name: 'no secret-bearing env var is client-exposed',
```
mattrossman, line 143: > `no secret-bearing env var is client-exposed` is expected to fail an agent that follows the page

Just noting this check is passing in my local runs and CI. Worth double checking if/why it's expected to fail as PR body mentions, in case there's something else we should be checking.

mattrossman (approved):
Looks good overall and tested a couple runs on my end.

I'd just call attention to [this potential flakiness](https://github.com/supabase/evals/pull/212#discussion_r3866814676) I experienced a couple times.

evals/build-docs-003-api-keys-guide/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+    /\.auth\s*\.\s*signUp\s*\(/.test(readText(file))
+  );
+
+  // A secret reaching Vite's client env is a leak whether or not this build
+  // happened to inline it, since these prefixes are exactly what gets inlined.
+  const prefixes = clientEnvPrefixes(root);
+  const exposed: string[] = [];
+  for (const file of files.filter((f) => /(^|\/)\.env/.test(f))) {
+    for (const line of readText(file).split('\n')) {
+      const [name, ...rest] = line.split('=');
```
barryroodt, line 117: This will let a client-exposed secret pass when `.env` contains `export VITE_ADMIN=sb_secret_example` and client code reads `import.meta.env.VITE_ADMIN`. Could you add that counterexample as a scorer fixture and make `no secret-bearing env var is client-exposed` reject it?

evals/build-docs-003-api-keys-guide/server.ts
```diff
@@ -0,0 +1,45 @@
...
+        notes:
+          'no server code under supabase/functions, so nothing proves the claim',
+      },
+    };
+  }
+
+  const offenders: string[] = [];
+  for (const file of files) {
+    const text = readText(file);
+    const hits = LEGACY_KEY_VARS.filter((variable) => text.includes(variable));
```
barryroodt: This will let a legacy server key pass when server code reads `ADMIN_KEY` and `supabase/functions/.env` binds that alias to the legacy key. Could you add that counterexample as a scorer fixture and make `server reads no legacy key variable` reject it?

evals/build-docs-003-api-keys-guide/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+
+function clientEnvPrefixes(hostWorkspace: string): string[] {
+  const config = readText(join(hostWorkspace, 'vite.config.ts'));
+  const declared = /\benvPrefix\s*:\s*(\[[^\]]*\]|['"`][^'"`]*['"`])/.exec(
+    config
+  )?.[1];
+  const prefixes = (declared?.match(/['"`]([^'"`]+)['"`]/g) ?? []).map(
+    (quoted) => quoted.slice(1, -1)
+  );
+  return prefixes.length ? prefixes : ['VITE_'];
```
barryroodt: This will let a client-exposed secret pass when `vite.config.ts` sets `const clientPrefix='PUBLIC_'; export default { envPrefix: clientPrefix }` and `.env` exposes it as `PUBLIC_ADMIN`. Could you add that counterexample as a scorer fixture and make the env guard reject it?

evals/build-docs-003-api-keys-guide/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  const offenders: string[] = [];
+  for (const file of clientSource) {
+    const hits = findSecrets(readText(file), status);
+    if (hits.length) offenders.push(relative(root, file));
+  }
+
+  // Static, because the scorer has no DOM to drive the built app through. It
+  // closes the hole where the publishable key is present but never used.
+  const signUpWired = clientSource.some((file) =>
```
barryroodt, line 105: This static check treats unreachable dead code containing `.auth.signUp(` as a successful sign-up, while an equivalent `registerUser()` wrapper fails without that literal token. Could you add both counterexamples as scorer fixtures and use a behavioral probe, or remove the static claim until the harness can observe the behavior?

barryroodt (commented):
Looking good, I left a few inline comments on scorer cases that can still pass when they should fail.

czenko:
**Update:** I added a couple more checks since Matt's review and have since reverted them since they proved to not reflect on the document but on the functionality of the CLI.

PR #216, opened 2026-08-19
fix(ci): merge the scheduled results PR with --admin

mattrossman:
## Problem

Sometimes (~15% of the time) nightly regression refreshes fail because of our required reviewers rule, leaving stale data in Hex.

For example last night's PR is stuck unmerged: https://github.com/supabase/evals/pull/215

You can see the corresponding failing step here: https://github.com/supabase/evals/actions/runs/32223257049/job/95980723901#step:13:1

CleanShot 2026-08-19 at 14 38 41@2x

## Fix

The bot is already on our bypass list, however it needs to opt-in with `--admin` which is similar to an AI team member clicking the checkbox on PRs:

CleanShot 2026-08-19 at 14 39 03@2x

Oddly, `gh` will happily allow the auto merge if the merge status is UNKNOWN which is why this wasn't noticed earlier. The flakiness is a race condition if `gh` has finished determining that the merge status is [blocked](https://github.com/cli/cli/blob/95d3a1db45abd547c2dafbee4f8a68ca53fb9c80/pkg/cmd/pr/merge/merge.go#L716-L733) before we attempt to auto merge. The `--admin` flag makes the bypass consistent.

Possibly related [issue](https://github.com/cli/cli/issues/13388) reporting the merge status quirk.

## How to review

We can't easily test this manually outside of the CI refresh path. So you'll have to just review based on the evidence [logged in our failing run](https://github.com/supabase/evals/actions/runs/32223257049/job/95980723901#step:13:1), and `gh` docs: https://cli.github.com/manual/gh_pr_merge

After merge, we should monitor the nightly refreshes and confirm they merge uninhibited.

## Note for the future

If we decide to add any additional required checks on PRs, we should add those on a separate ruleset, that bot's bypass ability only applies to the required reviewers check.

To prevent this footgun, I've renamed the ruleset to `main: review required` so its scope is clearer.

### DIFF
--- a/.github/workflows/eval-refresh.yml
+++ b/.github/workflows/eval-refresh.yml
@@ -505,6 +505,8 @@ jobs:
 
       - name: Merge scheduled results pull request
         # The app is on the ruleset's bypass list, so no review is required.
+        # --admin skips gh's own mergeStateStatus precheck, which doesn't account
+        # for bypass lists and intermittently refuses a merge the API allows.
         # Skips the merge when results are partial, so it stays open for review.
         if: >-
           github.event_name == 'schedule' &&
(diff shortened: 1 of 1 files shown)


mattrossman:
Planning to merge this today to avoid losing any more regression results

PR #218, opened 2026-08-19
DO NOT MERGE: point build-docs-003 at the api-keys content preview

czenko:
**Do not merge.** Scratch branch for measuring an unmerged docs revision. Every change here gets reverted, not landed.

Repurposed from the `build-docs-002` RLS preview experiment. That head is archived at `archive/rls-guide-preview-49276`, commit `b97c291`.

## Why

`build-docs-003-api-keys-guide` fails one check on most runs, and the failure is the same every time:

```
no legacy key reaches the server
  supabase/functions/roster/index.ts: SUPABASE_SERVICE_ROLE_KEY, SUPABASE_ANON_KEY
```

The published guide says to use the publishable and secret keys and that legacy keys are deprecated, then never shows how a key reaches code. No local development section, no environment variable convention, no code samples. An agent needing a server-side credential falls back on the Edge Functions convention it already knows, and that variable holds a legacy key.

The [content preview](https://docs-git-docs-api-keys-content-supabase.vercel.app/docs/guides/getting-started/api-keys) addresses that. It adds code samples, `NEXT_PUBLIC_` and `VITE_` prefix guidance, and names `SUPABASE_SECRET_KEYS` for server-side use. This branch points the eval at it.

## Changes

**`evals/build-docs-003-api-keys-guide/PROMPT.md`** — `REFERENCE` repointed to the preview. One line.

**`packages/core/src/docs-results.ts`** — `isSupabaseApexUrl` widened to accept `docs-git-*-supabase.vercel.app`. Without this a preview fetch is dropped from `docsCalls`, so `the agent read the API keys guide the prompt referenced` fails on a run that did read it. Lookalike hosts are still rejected and `docs-results.test.ts` passes unchanged, 53/53.

**`evals/build-docs-003-api-keys-guide/EVAL.ts`** — the guide-read check matches the preview hostname, not the path alone. Without this a fetch of the published page satisfies the check, and an agent that ignores the preview reads as having used it.

## What to look for

The signal is `no legacy key reaches the server`. It fails 4 of 6 runs against the published guide.

Read `the agent read the API keys guide the prompt referenced` first. If it fails, the agent never reached the preview and the rest of the run says nothing about the revision.

## Known limitation

The preview names `SUPABASE_SECRET_KEYS`, which the Edge Function runtime only injects on a newer CLI than this eval pins. An agent that follows that advice here gets an undefined value, which surfaces as a `roster returns every signed-up email` failure rather than a legacy-key failure. Read the two checks together.

## Running it

Label with `run-evals-changed`. Only `build-docs-003-api-keys-guide` changed, so that is the only eval that runs.

### DIFF
--- a/apps/web/src/data/regression-eval-results.json
+++ b/apps/web/src/data/regression-eval-results.json
@@ -351,7 +351,7 @@
       {
         "name": "no legacy key reaches the server",
         "passed": false,
-        "notes": "supabase/functions/roster/index.ts: SUPABASE_SERVICE_ROLE_KEY, SUPABASE_ANON_KEY"
+        "notes": "supabase/functions/roster/index.ts: SUPABASE_SERVICE_ROLE_KEY"
       },
       {
         "name": "roster returns every signed-up email",
@@ -373,25 +373,26 @@
         "supabase-postgres-best-practices"
       ],
       "loaded": [
-        "supabase"
+        "supabase",
+        "supabase-postgres-best-practices"
       ]
     },
     "docs": {
       "calls": [
         {
           "source": "web_fetch",
-          "query": "Explain the current Supabase API key system: what key types exist (publishable, secret, legacy anon/service_role), where each should be used (client vs server), how to fetch them via CLI or dashboard, and any security guidance about which keys must never be exposed to the browser.",
+          "query": "Summarize this page's guidance on Supabase API keys: what key types exist (anon, service_role, publishable, secret), how each should be used, and specifically any statements about which keys are safe to expose in client-side/browser code vs which must stay server-side only.",
           "hasContent": true,
           "pages": [
             {
-              "url": "https://supabase.com/docs/guides/getting-started/api-keys.md"
+              "url": "https://docs-git-docs-api-keys-content-supabase.vercel.app/docs/guides/getting-started/api-keys.md"
             }
           ],
-          "resultChars": 1757
+          "resultChars": 1348
         },
         {
           "source": "search_docs",
-          "query": "{ searchDocs(query: \"Edge Functions verify_jwt publishable key authorization header\", limit: 5) { nodes { title href content } } }",
+          "query": "{ searchDocs(query: \"edge function environment variables SUPABASE_SERVICE_ROLE_KEY secret key default secrets\", limit: 5) { nodes { title href content } } }",
           "hasContent": true,
           "pages": [
             {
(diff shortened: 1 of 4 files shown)


czenko:
Run [32307314965](https://github.com/supabase/evals/actions/runs/32307314965) complete: **41/43**, up from 38/43.

- `pgTAP test file(s) written under supabase/tests/` — now **passes**. The agent wrote 5 files, 31 assertions, covering all six seeded tables.
- `at least 8 assertions and all pass` — still fails, but on `28 passed, 3 failed` rather than `NOTESTS`. The floor is cleared; three assertions are wrong.
- Coverage judge — now runs and returns specific notes instead of "no test files to review".
- `the agent read the … guide` — passes, so the host-filter change did its job.

Full breakdown on [supabase/supabase#49276](https://github.com/supabase/supabase/pull/49276#issuecomment-5348737183).

This branch stays unmerged. The bot pushed a `chore: refresh eval results` commit onto it, which should not land either — the results are measured against a preview URL, not the published guide.

czenko:
This test branch fulfilled its purpose. Closing.

czenko:
**Result:** 43/43. First fully green run of this eval.

Verifies [#220](https://github.com/supabase/evals/pull/220), the reasoning-effort bump on the coverage judge, against the [#49276 preview](https://docs-git-docs-rls-tests-in-procedure-supabase.vercel.app/docs/guides/database/postgres/row-level-security.md). Run [32411045818](https://github.com/supabase/evals/actions/runs/32411045818).

**Setup:** only one variable moved against the nine baseline runs. Same preview URL, `skills: []` left in place, judge effort raised from `low` to `high`.

**Attempts:** the harness stops on pass, so `attempts: 2` means attempt 1 failed and attempt 2 passed.

| Attempt | Score | Test files | Six-table coverage |
| --- | --- | --- | --- |
| 1 | 41/43 | 3 | `list_members` and `list_items` appear only in setup inserts |
| 2 | 43/43 | 5 | dedicated files for both |

**Attempt 1 was a true red.** The rubric excludes setup that no assertion reads. Its suite asserted against four of the six tables, so a coverage failure is the correct verdict. That is agent variance between attempts, not judge variance.

**What this establishes:** the judge passed a complete suite and produced none of the false reds that made it the residual failure in the baseline runs.

**What it does not establish:** n=1. The baseline problem was the judge rejecting suites that were complete. One green run is a positive signal, not proof the variance is gone. Worth a second sample if that matters before merging #220.

**Not for main:** the bot pushed a `chore: refresh eval results` commit here. Those results are measured against a preview URL, so they should not land. Closing unmerged again.

czenko:
Closing again as I'm not running any tests.

czenko:
Closing. The check this branch was built to measure, `no legacy key reaches the server`, is removed from the eval in a0720f66. It scored which keys the Edge Function runtime hands a function, which the pinned CLI decides, so it reported the environment rather than the page.

The measurement did work: all six runs recorded a fetch of the preview host, which earlier attempts could not show because `docsCalls` dropped every non-`supabase.com` url. The `docs-results.ts` widening here is what made that visible and is worth keeping in mind next time an eval needs pointing at an unmerged docs PR.

On the nine checks that remain, the preview showed no regression against the published guide: 6/6 runs clean versus 5/6, where the single published failure was an agent that stopped early.

The RLS preview head this branch was repurposed from is archived at `archive/rls-guide-preview-49276`, commit b97c291.

PR #224, opened 2026-08-21
fix(evals): soften the pgTAP coverage rubric in the RLS guide eval

czenko:
Ref DOCS-1274

## Problem

`build-docs-002-rls-guide` measures whether the RLS guide produces good RLS. Tests are the bonus. The coverage judge graded test *craft*, and it was the eval's only unreliable check.

An eval passes only when every check passes, so one bonus check turned the whole doc red while all 40 RLS checks were green.

Four runs against the [#49276 preview](https://github.com/supabase/supabase/pull/49276):

| Sample | Judge config | Score | Judge verdict on audit |
| --- | --- | --- | --- |
| 1 | 5.5, high | 43/43 | Correct. Thorough suite, passed it |
| 2 | 5.5, high | 42/43 | Correct. `lives_ok` with no state checks |
| 3 | 5.6-sol, medium | 41/43 | Correct. Same, plus a broken assertion |

The judge was right every time. Its verdict tracked whichever suite the agent wrote that run, which varies. [#222](https://github.com/supabase/evals/pull/222) raising the default model and effort did not change that, because the variance was never in the judge.

## Solution

Keep the check. Scope the rubric to what the bonus is worth.

The bar is now one assertion against an application table, one allowed case, one denied case, and a role or identity set at least once. Ties go to a pass.

Every craft requirement that produced the reds is dropped, and each is named in the rubric as explicitly acceptable:

| Was required | Now |
| --- | --- |
| All six tables, by name | Covering only some is fine |
| Allow and deny for each of select, insert, update, delete | One allowed and one denied anywhere |
| Allowed write must be shown to change state | `lives_ok` alone is fine |
| Both `anon` and `authenticated` | One role is fine |
| Switch role and identity per case | At least once |
| `using`-filtered row asserted intact | Not required |

Renamed to `the pgTAP tests exercise access control on the application tables rather than standing in as placeholders`, since the old name claimed far more than the rubric now asks.

## What it still catches

A suite that tests nothing. Eight `select ok(true)` placeholders, assertions that never touch the application's tables, or a suite with no allowed case or no denied case anywhere. That floor is the reason to keep a judge here rather than delete the check.

## Manual testing

   ```bash
   pnpm typecheck && pnpm biome check evals
   ```

2. Label this PR `run-evals-changed` and confirm the coverage check passes on suites that previously failed it. Samples 2 and 3 above are the cases to beat: both were rejected only for `lives_ok` and partial coverage, which the rubric now accepts.

### DIFF
--- a/evals/build-docs-002-rls-guide/README.md
+++ b/evals/build-docs-002-rls-guide/README.md
@@ -32,4 +32,4 @@ Nothing in the migration says who may read or write what.
 
 Whether the agent arrives at pgTAP is itself a measurement. If agents do not write tests here, that says something about how reachable testing is from the guide.
 
-Coverage is judged against a low bar: one assertion on an application table, one allowed case, one denied case, and a role set at least once. Grading craft instead fails the whole eval on the bonus while every RLS check is green.
+Coverage is judged against a low bar: one assertion on an application table, one allowed case, one denied case, and a role set at least once. 
--- a/evals/build-docs-002-rls-guide/README.md
+++ b/evals/build-docs-002-rls-guide/README.md
@@ -32,4 +32,4 @@ Nothing in the migration says who may read or write what.
 
 Whether the agent arrives at pgTAP is itself a measurement. If agents do not write tests here, that says something about how reachable testing is from the guide.
 
-Coverage is judged, not pattern-matched. Two mechanical gates were tried and both failed in the same way: searching the file text counts a comment as a test, and searching the assertion descriptions rejects a correct suite that passes no description. Each fix in one direction made the other worse. The judge reads the sources with the six tables written into its rubric, and the suite's exit status proves the assertions ran.
+Coverage is judged against a low bar: one assertion on an application table, one allowed case, one denied case, and a role set at least once. Grading craft instead fails the whole eval on the bonus while every RLS check is green.
--- a/evals/build-docs-002-rls-guide/tests.ts
+++ b/evals/build-docs-002-rls-guide/tests.ts
@@ -50,11 +50,12 @@ export async function checkPgTapSuitePasses(
   };
 }
 
-export async function checkTestCoverage(
+/** A low bar on purpose: tests are the bonus here, so grading craft would fail the eval on the bonus while every RLS check is green. */
+export async function checkTestsExerciseAccessControl(
   ctx: LocalStackEvalContext
 ): Promise<CheckResult> {
   const name =
-    'tests assert allow and deny per operation on the seeded tables, for anon and authenticated';
+    'the pgTAP tests exercise access control on the application tables rather than standing in as placeholders';
   const files = await listTestFiles(ctx);
   if (files.length === 0) {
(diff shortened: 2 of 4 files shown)


evals/build-docs-002-rls-guide/tests.ts
```diff
@@ -50,11 +50,12 @@ export async function checkPgTapSuitePasses(
   };
 }
 
-export async function checkTestCoverage(
+/** A low bar on purpose: tests are the bonus here, so grading craft would fail the eval on the bonus while every RLS check is green. */
```
czenko:
```suggestion
/** A low bar on purpose: determining a great quality of tests is a tall order for low-effort agents */
```

evals/build-docs-002-rls-guide/README.md
```diff
@@ -32,4 +32,4 @@ Nothing in the migration says who may read or write what.
 
 Whether the agent arrives at pgTAP is itself a measurement. If agents do not write tests here, that says something about how reachable testing is from the guide.
 
-Coverage is judged, not pattern-matched. Two mechanical gates were tried and both failed in the same way: searching the file text counts a comment as a test, and searching the assertion descriptions rejects a correct suite that passes no description. Each fix in one direction made the other worse. The judge reads the sources with the six tables written into its rubric, and the suite's exit status proves the assertions ran.
+Coverage is judged against a low bar: one assertion on an application table, one allowed case, one denied case, and a role set at least once. Grading craft instead fails the whole eval on the bonus while every RLS check is green.
```
czenko:
```suggestion
Coverage is judged against a low bar: one assertion on an application table, one allowed case, one denied case, and a role set at least once. 
```

mattrossman:
I've tested by adding `run-evals-changed` to the PR, and removed / re-added it again for a second run. Note in those sample runs the agent didn't write pgTAP tests so they're still failing and not really testing the new softened rubric.

FWIW I don't think the eval failing is a problem! It's good to hold agents to a high standard, and I see this as an opportunity to strengthen our docs or skill guidance around pgTAP.

Eventually I think we can support [partial credit](https://linear.app/supabase/issue/AI-1114/support-partial-credit-in-scorers) in scorers to make the signal here more refined. In the meantime, if pgTAP truly isn't an important part of the guide, you might consider keeping the check for passive awareness but relaxing the all-or-nothing `passed: checks.every((check) => check.passed)` in the result.

Given that the sample run didn't test the updated rubric, let me know if you'd like to make this change anyway as a preventative measure or take another approach.

czenko:
> Given that the sample run didn't test the updated rubric, let me know if you'd like to make this change anyway as a preventative measure or take another approach.

@mattrossman Yes, I want to make this change anyway.

I have worked on the doc in an unmerged PR. That is why it is still failing. I have been running the Eval in a test branch and have verified that this passes whereas it was flakey and there was no more doc change I could do to make it pass. That's where this comes from.

mattrossman (approved):

PR #230, opened 2026-08-24
refactor(core): share the docs guide-read check across docs evals

czenko:
## Problem

`build-docs-002-rls-guide` builds its guide-read check by hand, and the API keys eval in #212 carries the same function again. Same filter, same `hasContent` split, same three-way notes string, same comment above it. The only difference is the guide's name.

Ten more docs evals are scoped. Each one points an agent at a page and needs to know whether the agent retrieved it, because without that a red run can't distinguish a guide that taught the wrong thing from a guide the agent never opened.

## Solution

- Adds `checkDocsGuideRead(toolCalls, { path, label })` to `docs-results.ts`, beside `buildDocsResult`. It returns the `CheckResult` the eval-local copies were building.
- Points `build-docs-002-rls-guide` at it and deletes the local copy.
- Keeps the check name string byte-identical, so result history doesn't break.
- Covers the three outcomes with tests: guide retrieved, guide reached with no page content, guide not reached.

## Manual testing

1. Run `pnpm --filter @supabase-evals/core test`. 56 pass, including three new `checkDocsGuideRead` cases.
2. Run `pnpm typecheck`. Clean. It covers `evals/**/EVAL.ts`.
3. Read the `build-docs-002-rls-guide/EVAL.ts` diff. Confirm the name passed to `label` reproduces the removed check's name exactly.

### DIFF
--- a/packages/core/src/docs-results.ts
+++ b/packages/core/src/docs-results.ts
@@ -369,9 +369,6 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
  * Whether the agent retrieved the guide at `path`, and by what route. A
  * search_docs hit carries the url in its result, not its request, so this
  * reads the harness's resolution rather than raw tool calls.
- *
- * Nothing here reads the prompt. The caller owns passing a path its
- * PROMPT.md points at.
  */
 export function checkDocsGuideRead(
   toolCalls: ToolCallRecord[],
--- a/packages/core/src/docs-results.ts
+++ b/packages/core/src/docs-results.ts
@@ -1,4 +1,9 @@
-import type { DocsCall, DocsCallPage, DocsResult } from './eval-metadata.js';
+import type {
+  CheckResult,
+  DocsCall,
+  DocsCallPage,
+  DocsResult,
+} from './eval-metadata.js';
 import type { ToolCallRecord } from './index.js';
 import { isRecord } from './json.js';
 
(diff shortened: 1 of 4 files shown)


packages/core/src/docs-results.ts
```diff
@@ -359,3 +364,33 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
...
+ * Whether the agent retrieved a guide the prompt pointed it at, and by what
+ * route. A search_docs hit carries the guide's url in its result rather than
+ * its request, so this reads the harness's own resolution instead of scanning
+ * raw tool calls.
+ *
+ * Docs evals score the page's effect on what the agent built. Without this
+ * check a red run is ambiguous: the guide may have taught the wrong thing, or
+ * the agent may never have opened it.
+ */
+export function checkDocsGuideRead(
```
czenko, line 373: Worth noting, since this is an architectural decision: this is the first `CheckResult` builder in core. Core has parsed traces into data and left the pass/fail decision to evals until now.

packages/core/src/docs-results.ts
```diff
@@ -359,3 +364,31 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
...
   return { calls };
 }
+
+/**
+ * Whether the agent retrieved the guide at `path`, and by what route. A
+ * search_docs hit carries the url in its result, not its request, so this
+ * reads the harness's resolution rather than raw tool calls.
+ *
+ * Nothing here reads the prompt. The caller owns passing a path its
+ * PROMPT.md points at.
```
czenko:
```suggestion
```

czenko:
Parking this as a draft per the thread. #212 is unstacked and standalone on `main` now, carrying its own copy of the check.

Holding until the structured docs-injection idea is decided. If a `includeDocs:` frontmatter field lands, this check goes vacuous and this PR should just close. If it doesn't, this is still the cleanup once a third docs eval needs it.

PR #240, opened 2026-08-25
fix(sandbox): install skills natively per harness (AI-1034)

seanoliver:
## What's inside

- ~215 lines of skills install changes ([skills.ts](https://github.com/supabase/evals/pull/240/changes#diff-badfb2869f8854f19b00a8e02d6e9d94ef4e1cea66004399b290c03b8ecfb1fa), [local-stack-runtime.ts](https://github.com/supabase/evals/pull/240/changes#diff-366863d3b55e1541a2a038a7af0efb7952e345a84cdcdcce85727ca5dd0aef7f))
- ~5 lines of plumbing that does nothing yet ([index.ts](https://github.com/supabase/evals/pull/240/changes#diff-9a4ceebe7c6f86856371906c3f061d3b56b7457022b05179884a113e7ced67e8)), for the PR stacked on this one
- ~195 lines of tests, skimmable ([unit.test.ts](https://github.com/supabase/evals/pull/240/changes#diff-d9864aa0981cd32280dc8a1cca41ada618e91ef3684f52c4c49234e378aee648), [docker.test.ts](https://github.com/supabase/evals/pull/240/changes#diff-216a9cc0506fc98aac7b13fd029e70581f3e089c35dcde73fac0f2217e93b00f))

Bottom of a 2-PR stack under #241, which removes the system prompt we write for the CLI agents.

Claude Tag wrote the first pass at this (previously https://github.com/supabase/evals/pull/180, now closed).

## Problem

Installing a skill means copying its folder somewhere an agent will look, and each agent reads a different directory.

- `skills add` takes a flag naming which agents to install for, but we weren't not passing it.
- Without that flag, the CLI tries to detect an installed agent, and when it cannot, it installs for all 71 it knows about.
- So every sandbox got ~52 folders it had no use for: `.aider-desk`, `.factory`, `.kilocode`, `.windsurf`, `.zencoder`, and non-dotted `data/` and `skills/` among them. Each holds a copy of the same two Supabase skills.
- It also didn't install Codex skills into the native `.agents/skills` folder, which required us to include an addendum to the Codex prompt to get it to look in the Claude skills folder (`.claude/skills`).

### Install

Name the three agents we run in this repo:

```bash
skills add '<dir>' --agent claude-code codex opencode --skill '*' --copy --yes
```

That writes two directories, `.claude/skills` and `.agents/skills`, instead of 52.

`--copy` is required so that `.claude/skills` gets created in an empty eval workspace. Without it only `.agents/skills` shows up.

After installing, `installSkills` confirms every skill is actually present in both directories (`.claude/skills` and `.agents/skills`). It used to check `.claude/skills` only.

### Plumbing for the next PR

The prompt builders now take the agent as an argument and ignore it. This is used by #241 to skip the prompt text for the CLI agents.

## How to review

Confirm the new install works:

```bash
pnpm --filter @supabase-evals/sandbox test:docker
```

If interested, run both install strategies side by side in a temp folder to see the difference:

```bash
mkdir -p /tmp/probe/stage/skills/demo /tmp/probe/old /tmp/probe/new
printf -- '---\nname: demo\ndescription: Demo.\n---\n\nBody.\n' > /tmp/probe/stage/skills/demo/SKILL.md

# old, no --agent
cd /tmp/probe/old
env -u CLAUDECODE -u CLAUDE_CODE_ENTRYPOINT npx -y skills@1.5.11 add /tmp/probe/stage --skill '*' --copy --yes
ls -A

# new, three agents named
cd /tmp/probe/new
env -u CLAUDECODE -u CLAUDE_CODE_ENTRYPOINT npx -y skills@1.5.11 add /tmp/probe/stage --agent claude-code codex opencode --skill '*' --copy --yes
ls -A
```

You should see 53 entries (old) vs. 3 (new).

## Follow up tasks

- Remove `skills-lock.json` from the workspace after installing. The CLI drops it in the root and it ends up in the exported workspace, same as before this change.
- Assert the full set of workspace directories in the docker test. It currently samples five of the ~52 the old fallback created, so it catches that fallback firing but does not prove nothing else was written.
- Add a check that each agent actually surfaces the skills it finds. Both this PR and #241 assume it, and nothing tests it.

Ref AI-1034


packages/sandbox/src/skills.ts
```diff
@@ -1,34 +1,65 @@
...
+ *
+ * Naming them explicitly also matters. With no `--agent` flag the CLI falls
+ * back to *every* agent it knows when it cannot detect an installed one,
+ * littering the scored, exported workspace with ~52 stray entries. That
+ * fallback does happen to cover both directories we want, so it is not why
+ * skills reach an agent; it is a workspace-pollution and determinism problem.
+ * Detection depends on the surrounding environment, so naming the agents is
+ * what makes the install predictable.
+ */
+export const SKILLS_INSTALL_AGENTS = [
```
mattrossman: When someone tries to add a new harness elsewhere in the repo, can we make this fail loudly if they forget to update it, instead of silently running evals w/o skills?

An exhaustive `Record<AgentHarnessId, string | null>` could do it, since the [runner relies](https://github.com/supabase/evals/blob/2252b77ededab0fabc6562b222d1560d6d7df38a/packages/core/src/agents/types.ts#L80-L81) on the same id. `SKILLS_INSTALL_AGENTS` and `SKILLS_INSTALL_DIRS` could then be derived from it.

E.g.

```ts
const SKILLS_PATH_BY_AGENT: Record<AgentHarnessId, string | null> = {
  'ai-sdk': null,
  'claude-code': '.claude/skills',
  codex: '.agents/skills',
  opencode: '.agents/skills',
};

export const SKILLS_INSTALL_AGENTS: AgentHarnessId[] = [];
export const SKILLS_INSTALL_DIRS: string[] = [];
for (const id of agentHarnessIdSchema.options) {
  const path = SKILLS_PATH_BY_AGENT[id];
  if (path === null) continue;
  SKILLS_INSTALL_AGENTS.push(id);
  if (!SKILLS_INSTALL_DIRS.includes(path)) SKILLS_INSTALL_DIRS.push(path);
}

```

seanoliver: Done: `SKILLS_PATH_BY_AGENT` is the single source and both lists derive from it. A new harness id won't compile until its scope is declared, and the error names the missing harness.

`SKILLS_INSTALL_DIR` reads from the same map, so the prompt listing can't drift from the install.

mattrossman: Thanks, confirming I now get a type error if adding a harness w/o specifying the skills path!

CleanShot 2026-09-07 at 11 31 00@2x

README.md
```diff
@@ -163,7 +163,7 @@ To use a skill in an experiment, reference its directory name in the experiment'
...
 
-- **Local-stack (sandbox) mode:** skills are installed into the workspace with [Vercel's `skills` CLI](https://github.com/vercel-labs/skills) (baked into the sandbox image, sourced from the local `skills/` directory — never the network) under `.claude/skills/`. When a task matches, the agent reads `.claude/skills/<name>/SKILL.md` (and any files it references) with its file tools.
+- **Local-stack (sandbox) mode:** skills are installed into the workspace with [Vercel's `skills` CLI](https://github.com/vercel-labs/skills) (baked into the sandbox image, sourced from the local `skills/` directory — never the network), into every project scope the CLI harnesses discover natively: `.claude/skills/` for Claude Code, `.agents/skills/` for Codex and OpenCode. Each CLI discovers the scope it reads and surfaces those skills to the model itself. The framework also still injects a listing naming them, so a CLI harness hears about them twice; removing that is a separate change.
```
mattrossman, line 166: nit, I'd avoid temporal comments about "a separate change" in the user-facing README. Even though it's reverted in the subsequent PR, I try to keep the trunk of a PR stack mergable on its own based on [this](https://supabase.slack.com/archives/C0161K73J1J/p1786970385765769?thread_ts=1786969914.393969&cid=C0161K73J1J) guidance.

mattrossman (approved):
Sanity checked `pnpm eval -- --eval build-cli-002-declarative-schema --experiment codex-gpt-5.4-mini`

In that run it was reading Claude's skill files instead but it sounds like that will be fixed by subsequent PR #241 in the stack by trimming the `buildSkillsPrompt`

PR #229, opened 2026-08-24
feat: average eval scores across independent runs

mattrossman:
## What's inside

- ~560 lines of framework changes (e.g. [run-eval.ts](https://github.com/supabase/evals/pull/229/changes#diff-0d364314dfec4f47eb26a906c4c5bf60351fefa57d8449a93d0d03e2466c1434), [export-results.ts](https://github.com/supabase/evals/pull/229/changes#diff-d781f8435c9ec3843c38fce9cc1e1a5f657d88fb4951fc2e5dfcd84064cecd56))
- ~120 lines of CI / Vercel Sandbox orchestration (e.g. [eval-refresh.yml](https://github.com/supabase/evals/pull/229/changes#diff-943b3ca0618de4aae9edbfa6d5f40c7409050c29c6802f3a9752ebde3d76a03b), [run-vercel-evals.ts](https://github.com/supabase/evals/pull/229/changes#diff-8e35b802b923520e05500a9549c95311442f6b6d37ec79db609713fec57f10f1))
- ~300 lines of UI changes (e.g. [selection-sheet.tsx](https://github.com/supabase/evals/pull/229/changes#diff-4d44837bdc2575f5f362ca9ab63347f9f0d6afe3abea2d48eb66e344d00cf2c7), [eval-details.tsx](https://github.com/supabase/evals/pull/229/changes#diff-71474b683db8b9ee250a2affa513bea584dbf0b6add8c8749feafc1e055ea5f1))
- ~245 lines of tests, skimmable

I intentionally omitted results from the final diff and made UI changes backwards compatible so we can do the refresh in a follow-up PR, to reduce review surface area.

I considered stacked PRs, but there's not a clean way to incrementally add just the framework changes without breaking our `--runs` interpretation or adding pointless backwards compat temp fluff.

## Problem

Runs aren't sufficiently reliable. We assign a binary `Pass`/`Fail` on each `experiment x eval` pair, which means results can shift significantly across refreshes.

Today we partly denoise failures by giving agents 1 extra attempt upon failure, but that biases our results and forces more work in serial (since runs can be conditional on the prev running passing or not).

## Changes

Each `experiment x eval` pair now runs multiple times, in parallel (in CI). Website averages results across runs. Both benchmark and regression results use this strategy.

I'm picking 3 runs because it's the smallest starting point that gives us tie breakers. Based on how CI handles the significant increase in Sandboxes, we can re-evaluate from there.

### Framework
[...]

Issue #164, opened 2026-08-04
Base prompt


JesseVent:
Were any evals ran on the base prompt itself? 

Saying it's solving a 'Supabase eval task' potentially has the following impacts:
- nudges it toward Supabase tools instead of generic ones
- leaks context a user would never provide
- each mode gets a differently worded prompt
- instructing the AI to end its turn with a short summary 

Could shape how long it runs and when it stops which also could skew comparisons.

You can view the trace here https://supabase-evals-two.vercel.app/?trace=build-storage-001-private-bucket-access

https://github.com/supabase/evals/blob/0f19647187097add479e7146dc3c5a7504322c64/apps/framework/harness/run-eval.ts#L309-L322

Could be better to just leave it default.

You can view the trace here https://supabase-evals-two.vercel.app/?trace=build-storage-001-private-bucket-access

<img width="1766" height="1039" alt="Image" src="https://github.com/user-attachments/assets/4d3e2903-5e3f-41b9-a26c-fe8fe8c09a36" />

Trace from one of the failed ones #161 
<img width="1796" height="1031" alt="Image" src="https://github.com/user-attachments/assets/c0e7383e-59f0-453c-9044-7b40fbf3bad9" />

mattrossman:
Thanks for surfacing this @JesseVent, we've removed those references in https://github.com/supabase/evals/pull/241 along with the custom skill loading instructions leftover from the AI SDK harness in https://github.com/supabase/evals/pull/240 so that each terminal agent gets more of a clean slate resembling the out-of-box experience. The next time we refresh results it'll include those changes.

PR #241, opened 2026-08-25
fix(framework): remove the harness system prompt for CLI agents (AI-1034)

seanoliver:
This is stacked on #240. Review that first; this diff is against it. No results are touched.

## Problem

We wrote our own system prompt into every eval run, for all four agents:

> You are an agent solving a Supabase eval task in a Linux workspace. Use the provided tools to inspect and modify the workspace and run commands. When you are done, end your turn with a short summary of what you did.

- Claude Code, Codex and OpenCode ship their own system prompt. Ours was layered on top, and Codex and OpenCode have no system-prompt flag, so for them it went onto the user prompt.
- It described `bash` and `files_read`, which only the in-process `ai-sdk` agent has.
- "End your turn with a short summary" shapes `agentReport`, which the LLM judge reads. That is a scoring change.
- It says Supabase, and "project". An agent that would otherwise pick a database has been handed the answer (#164).

## Changes

| Agent | Before | After |
|---|---|---|
| `ai-sdk` | framing + tool list + skills list | one neutral line + tool list + skills list |
| `claude-code` | ours appended to its own | nothing |
| `codex` | ours prepended to the user prompt | nothing |
| `opencode` | ours prepended to the user prompt | nothing |

- CLI agents take no system prompt at all. `createCliAgent` no longer stages a system-prompt file or passes a path to the runner; `--append-system-prompt-file` and the Codex/OpenCode prepend are gone. A non-empty `systemPrompt` for a CLI agent throws.
- The `ai-sdk` base prompt is now `You are an agent. Use the provided tools to complete the task.` It is the same in both modes. No Supabase, no project, no stopping instructions.
- `buildToolSurfaceAddendum` and `buildSkillsPrompt` return `''` for every CLI agent. The bare sandbox no longer takes an `agent` or returns a `promptAddendum`.
- Prompt assembly moves out of `run-eval.ts` into `system-prompt.ts` so it can be unit tested.
- `run-eval.ts` only starts a run when it is the entry point. It used to call `main()` on import, which the harness unit tests now trigger.
- Each run records `systemPrompt` in its raw `result.json`. `export-results` drops it, so nothing reaches the site.

Since the first review: the system-prompt-file plumbing for CLI agents is deleted rather than made optional, and the ai-sdk base prompt is rewritten for #164. The branch is re-applied on the current #240 (the retry loop is gone there), so the commits are new. #240 now has `main` merged in, which brings the usage and duration fields from #270.

## Testing

- `pnpm --filter @supabase-evals/core test`, `pnpm --filter @supabase-evals/sandbox test`, `pnpm --filter @supabase-evals/framework test`, `pnpm --filter @supabase-evals/web test`: all pass locally.
- `typecheck` for framework, sandbox and web passes. `biome check .` is clean.
- Unverified against real CLIs since the deletion. The earlier sandbox runs (Claude Code [32785480269](https://github.com/supabase/evals/actions/runs/32785480269), Codex [32880822811](https://github.com/supabase/evals/actions/runs/32880822811), OpenCode [32880839623](https://github.com/supabase/evals/actions/runs/32880839623)) recorded an empty prompt and still loaded the Supabase skills, and the path they exercised (no flag, task alone) is the one that remains.

## Score movement

Full benchmark on both branches before this revision, artifact only, nothing published. Both used `runs=2`.

| | #240, no prompt change | #241, prompt removed | |
|---|---|---|---|
| all pairs | 86.0% | 84.7% | 21 flips, 12 down 9 up |
| `claude-code` | 92.3% | 92.3% | no change |
| `codex` | 81.5% | 75.0% | down 6.5 points |
| `opencode` | 82.6% | 89.1% | up 6.5 points |

This looks like run-to-run variability. Once #229 lands and scores average across runs, the numbers should settle.

## Follow-ups

- Re-run the benchmark comparison once #229 lands and the numbers can carry attribution.
- Drop the "end your turn with a short summary" line from `evals/resolve-performance-001-slow-query-cpu-spike/PROMPT.md:15`, the last one left. It sits in the task text so it hits every agent equally, and removing it invalidates that eval's recorded results, so it should be a separate change.

Ref AI-1034


mattrossman:
Tested running `pnpm eval -- --eval build-cli-002-declarative-schema --experiment codex-gpt-5.4-mini` and I see Codex reading its natively supported `.agents/skills` files now instead of Claude's.

CleanShot 2026-09-07 at 12 57 35@2x

apps/framework/harness/system-prompt.ts
```diff
@@ -0,0 +1,76 @@
...
+ * Base framing for the ai-sdk harness: what it can't infer on its own — that it
+ * has tools, and what they act on. Deliberately silent on how to finish a turn
+ * (no "end with a summary"): stopping behaviour is part of what's measured.
+ * Empty for every CLI harness.
+ */
+function basePromptFor(agent: AgentHarnessId, mode: EvalMode): string {
+  if (agent !== 'ai-sdk') return '';
+  if (mode === 'local-stack') {
+    return (
+      'You are an agent solving a Supabase eval task in a Linux workspace. ' +
```
mattrossman: Could we also tie in the fix for https://github.com/supabase/evals/issues/164 here? I.e. we shouldn't mention that this is a Supabase eval or that there is a "project" needing to be modified, as that could give an unfair advantage to the non terminal-agent experiments.

I recognize it's a minor issue at this point, since only the AI SDK harness is affected here, which is no longer used in our benchmarks / regression runs. Might end up taking it out entirely. But also easy to clean up while we're here.

packages/core/src/agents/engine.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  return { systemPromptPath, commands };
+}
+
+describe('createCliAgent prompt staging', () => {
+  // The engine requires the runner's API key before it stages anything.
+  beforeEach(() => {
+    process.env[API_KEY_ENV_VAR] = 'k';
+  });
+
+  it('stages a system prompt and hands its path to the runner', async () => {
```
mattrossman: Do we still need to support system prompt overrides for terminal agents?

Seems like that's now legacy behavior we won't want to support going forward, I think there's a cleanup / code deletion opportunity here. It's a bit convoluted to support the plumbing of writing a system prompt to a file that's read from inside the container.

The only use case I can think of is the executor example cited in the PR, but (a) the only experiments that use executor are AI SDK ones which don't need this system prompt file workaround (b) we don't use the executor experiments, they're just POC placeholders (c) we could achieve the same thing with an addendum on the user prompt.

apps/framework/harness/run-eval.ts
```diff
@@ -367,6 +343,13 @@ async function runOne(
     transcript: TranscriptPart[];
     agentReport: string;
     stoppedReason: string;
+    /**
+     * The exact system prompt handed to the agent (`''` when it got none). CLI
```
mattrossman: If we end up keeping this for terminal agents, I'd clarify here that it's more of an optional thing [appended](https://github.com/supabase/evals/blob/5bed078b387b2cd9807ff9179e019bd2ba42f998/packages/core/src/agents/claude-code/runner.ts#L70-L76) to the system prompt, not replacing it with this exact string alone.

apps/framework/harness/system-prompt.ts
seanoliver: Done. The base prompt is now one line, the same in both modes: `You are an agent. Use the provided tools to complete the task.` No Supabase, project, or eval in it, and a test asserts that.

packages/core/src/agents/engine.test.ts
seanoliver: Agreed. It's deleted: `createCliAgent` no longer stages a system prompt file or passes a path to the runner, and the three runners lost their flag/prepend branches. A non-empty `systemPrompt` for a CLI agent now throws instead of being dropped, so pairing the executor MCP with a CLI agent fails loudly. If anyone wants that pairing later, the protocol text goes on the user prompt as you suggested.

apps/framework/harness/run-eval.ts
seanoliver: This is moot now that the override is gone. What remains here is that each run records the `systemPrompt` it sent, `''` for CLI agents.

mattrossman (approved):

PR #261, opened 2026-09-03
feat(evals): eval the server-side auth client guide in supabase.com/docs

czenko:
Closes DOCS-1303

## Problem

`auth/server-side/creating-a-client` carries the highest agent share of any task-driven page in the docs cohort, 35.0 percent on 22,918 views, and `Server-side Auth` is the densest label in feedback intake. Nothing measures whether the page gets an agent to a dashboard that will not hand itself to a forged cookie.

## Solution

- Adds `build-docs-005-server-side-auth`, `suite: regression`, `services: [gotrue, kong, postgrest]`.
- Seeds a Next.js 16 app: a sign-in form posting to `POST /login`, an unwired login route handler, and an unwired dashboard. The seed fixes the form field names, the login contract, and `data-testid="viewer-email"` on the dashboard.
- **The one claim is measured behaviorally.** `a tampered session cookie does not get the dashboard` signs a user in through the app, re-encodes the session with a second user's identity, and leaves the signature alone so the token no longer verifies. Reading stored session state accepts that cookie; verifying the token rejects it. It is gated on the render control passing first, so an app that renders nothing cannot clear it.
- Five runtime checks install, build, and serve the app inside the sandbox with `ctx.exec`. `next` is not a framework dependency, and the host-side build helpers relink the workspace's `node_modules` to the framework's own, so a host-side build cannot see what the seed declared.
- Three source checks cover the verifying call, the cookie adapter, and the deprecated helpers import.
- Checks are sourced from outside the page: feedback intake (FDBKIN-20970, FDBKIN-6870, FDBKIN-10628, FDBKIN-14804, FDBKIN-15214, FDBKIN-11678, FDBKIN-4573), Supacademy `auth/troubleshooting/02-session-refresh-and-ssr-failure-patterns.md`, and `examples/prompts/nextjs-supabase-auth.md` in the main repo.

**The verifying-call check accepts `getClaims()` and `getUser()` both.** DOCS-1303 and the scoping doc ask for `getUser()`; the page now recommends `getClaims()` and names it the way to protect pages. A check on either alone would report the page as broken for giving current advice, so the check names the class.

Two framework problems this surfaced, neither fixed here:

- `ctx.stackStatus()` requires `API_URL`, `PUBLISHABLE_KEY` and `SECRET_KEY` together and throws when any is missing. This eval needs the api url and one client key, so it reads `supabase status` directly and accepts `PUBLISHABLE_KEY` or `ANON_KEY`.
- `discoverEvals` reads `PROMPT.md` in every directory under `evals/` and throws when one is absent, so a git-excluded `solutions/` directory left behind by a branch switch breaks discovery for every eval in the repo.

## Baseline

**What ships:** three runs in `docs-eval-results.json`, codex `gpt-5.6-luna`
without skills. Each scores 8/9.

**The failing check is the guide-read check.** `web_search` reached the page and
returned no page content, so the check cannot confirm the page was opened. That
is a harness instrumentation gap, not a finding about the page. The other eight
checks pass on all three runs, including the tampered cookie check.

**What this says:** read it as regression cover rather than benchmark signal. The
page gets the agent to a dashboard that rejects a forged cookie. It does not say
the page caused that, because a model that already knows the `@supabase/ssr`
pattern produces the same output. Hardening is worth considering after DOCS-1313
lands, so the two changes do not confound each other.

## Manual testing

1. Confirm the ports are free.

   ```
   docker ps --format '{{.Names}}\t{{.Ports}}' | grep -E '5432[12]->'
   ```

2. Confirm discovery and frontmatter.

   ```
   pnpm eval:dry -- --eval build-docs-005-server-side-auth
   ```

   One `PLAN` line per experiment, each reading `stage=build suite=regression mode=local-stack`.

3. Typecheck.

   ```
   pnpm typecheck
   ```

   Clean.

4. Lint the eval.

   ```
   ./node_modules/.bin/biome check evals/build-docs-005-server-side-auth
   ```

   `No fixes applied.`


nrichers (approved):
LGTM. There's a nit below but it doesn't seem hugely relevant.  🚀 

---
Verdict: APPROVE WITH NITS
Type: new eval (docs suite) — evals/docs/build-docs-005-server-side-auth

TL;DR: Tests whether agents build server-side auth checks using getClaims/getUser
rather than trusting getSession() alone. Motivation cites 7 FDBKIN tickets plus a
35% agent-share / 22,918-views metric. Scorer genuinely forges a JWT (re-signs
claims, invalidates signature) and asserts the app rejects the forged token while
accepting a real cookie — a real security-behavior check, not a superficial one.

Checks:
- motivation: field — concrete, multi-ticket
- Prompt casualness — no auth vocabulary (cookie/session/token/verify) leaked in,
  per README's own self-check
- Scorer structure — one flat checks[] array (probes + source-pattern regex +
  guide-read), no branching accumulation
- Security probe validity — verified in probes.ts and source.ts: real cookie-
  adapter shape, correct getClaims/getUser vs getSession-only detection
- Refreshed results — docs-eval-results.json, commit 0606db0
- pnpm typecheck — pass (via node@24 override)
- pnpm format:check — pass (2 pre-existing unrelated symlink warnings)
- gh pr checks — pass (only Vercel preview + codesmith; no lint/build gate)

Findings (new this pass):
- PR body claims "six runs, sonnet-5 with and without skills, 9/9 on all six" —
  zero sonnet-5 entries exist in any committed results file
  (eval-results.json, regression-eval-results.json, docs-eval-results.json).
  That claim traces to local/ad-hoc testing, not what actually ships in the PR.
  Worth correcting the PR body so reviewers aren't misled by data that isn't
  there.

Nits (not blocking):
- All 3 committed runs show overall passed: false, but only on "did the agent
  read the guide" (web_search returned no page content) — a harness/tool-
  choice artifact, not a real doc failure. The 3 actual security/behavioral
  checks pass 3/3. Worth a follow-up note, not a blocker.

Blockers: none.

PR #263, opened 2026-09-04
chore(platform-lite): bump Supabase MCP to 0.12.0

barryroodt:
## Summary

- Bump the platform-lite Supabase MCP dependency to 0.12.0.
- Refresh the pnpm lockfile, including transitive mcp-utils 0.8.0.
- Keep the platform-lite integration unchanged after reviewing the 0.12.0 cost-confirmation behavior.

## Changelog review

platform-lite uses the MCP platform API with an access token and API URL. It does not enable cost confirmation or depend on the legacy cost tools, so 0.12.0 needs no integration change.

## Testing

- `pnpm install --frozen-lockfile`
- `pnpm --filter @supabase-evals/platform-lite typecheck`
- `pnpm exec biome check packages/platform-lite`
- `pnpm --filter @supabase-evals/platform-lite exec vitest run test/mcp-platform.test.ts test/clickhouse-logs.test.ts`

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1253,7 +1253,7 @@ async function getAvailablePort(): Promise<number> {
 }
 
 export const ACCESS_TOKEN = 'eval-token';
-export const MCP_SERVER_VERSION = '0.11.0';
+export const MCP_SERVER_VERSION = '0.12.0';
 // Well-formed but inert PAT used when a Supabase MCP server is docs-only: the
 // server requires a token to boot but never authenticates without a platform.
 const THROWAWAY_ACCESS_TOKEN = `sbp_${'0'.repeat(40)}`;
--- a/packages/platform-lite/package.json
+++ b/packages/platform-lite/package.json
@@ -28,7 +28,7 @@
     "zod": "catalog:"
   },
   "devDependencies": {
-    "@supabase/mcp-server-supabase": "^0.11.0",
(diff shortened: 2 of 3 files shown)


Rodriguespn (changes requested):
Thanks for take care of this Barry. AFAIK, we would still need to bump the MCP server version in here right?
https://github.com/supabase/evals/blob/1d859a7ff9bf423648bf427711725871e90cea82/packages/core/src/index.ts#L1256

This is the MCP Server version we're using when running the evals. 

```bash
npx @supabase/mcp-server-supabase@0.11.0 --access-token <token> --features docs,account,database,development,debugging,functions [--api-url <url>]
```

The npm package used in `packages/platform-lite/package.json` is used only on the platform-lite tests

barryroodt:
Good catch @Rodriguespn. I bumped the eval launcher default to 0.12.0 in [`efd8bc8`](https://github.com/supabase/evals/commit/efd8bc8c71de40b97bec08c1897d31d5d5a06bc2).

Rodriguespn (approved):
Approved 🚢  
Platform-lite bumps to `^0.12.0` and the harness runtime pin (MCP_SERVER_VERSION) now tracks `0.12.0`; lockfile clean, typecheck and all core/platform-lite tests pass, and no integration change is needed (cost confirmation is opt-in and unused).

PR #269, opened 2026-09-07
feat: replace GPT-5.4-mini experiments with GPT-5.6 Luna

Rodriguespn:
## What

Replaces the four GPT-5.4-mini experiments with GPT-5.6 Luna, the new Assistant default model, so the model change in supabase/supabase#49749 is covered by the public eval suite.

## Why

The Assistant default model changed to `gpt-5.6-luna` (supabase/supabase#49749); the public evals suite should track that model.

Closes AI-1190


mattrossman (approved):

experiments/codex-gpt-5.6-luna-no-skills.ts
```diff
@@ -9,7 +9,7 @@ import { localStackRuntime } from '@supabase-evals/sandbox';
 export default defineExperiment({
   suite: ['no-skills'],
   agent: codexAgent({
-    model: 'gpt-5.4-mini',
+    model: 'gpt-5.6-luna',
```
mattrossman, line 12: Could we rename the other 5.6 experiment files while we're here? It's not obvious from reading them what `gpt-5.6` means next to `gpt-5.6-luna`, I think `gpt-5.6-sol` would be clearer.

Image

Rodriguespn: Nice catch. Done ✅

PR #273, opened 2026-09-08
feat: add @supabase/middleware regression eval

mandarini:
Adds a regression eval for `@supabase/middleware` (https://github.com/supabase/middleware), the sibling of the `@supabase/server` pair: the prompt names the package and asks for an Edge Function that composes the built-in CORS middleware with a hand-written `defineMiddleware` API-key gate. The scorer drives the key gate through the local gateway (no key and wrong key both `401`, right key `200` with the key id read from `ctx`), checks `Vary: Origin` as the CORS signal, and gates on the source importing the package and calling `defineMiddleware`.

Two details worth knowing: the local Kong answers preflights and rewrites `Access-Control-Allow-Origin` to `*` on `/functions/v1/*`, so `Vary: Origin` is the only CORS behavior observable through it; and the seeded `functions/.env` needed a `.gitignore` exception, following the existing one. Refreshed results are included: 6 of 6 runs pass (three with skills, three without), the expected baseline for a usability canary.

### DIFF
--- a/evals/build-functions-007-cors-api-key-with-middleware/EVAL.ts
+++ b/evals/build-functions-007-cors-api-key-with-middleware/EVAL.ts
@@ -1,0 +1,207 @@
+import {
+  type CheckResult,
+  type LocalStackEvalContext,
+  type LocalStackScorer,
+} from '@supabase-evals/core';
+
+// Regression eval for @supabase/middleware: the prompt names the package and
+// asks for one built-in (CORS) plus one hand-written middleware (an API-key
+// gate) composed in front of a handler. It guards the usability property —
+// "when pointed at the package, agents compose with it correctly" — as a
+// canary for package or docs changes that break agent usage. The "uses
+// @supabase/middleware" and "uses defineMiddleware" checks are GATING.
+//
+// Requests go through the local Kong gateway, which owns CORS on
+// /functions/v1/*; see README.md for what that leaves observable.
+const FUNCTION = 'notes-api';
+const ORIGIN = 'https://app.example.com';
+// Mirrors local/supabase/functions/.env, which the agent is told about.
+const API_KEY = 'nk_live_7f3a9c';
+const KEY_ID = '7f3a9c';
+
+interface InvokeResult {
+  status: number;
+  ok: boolean;
+  body: string;
+  headers: Headers;
+}
+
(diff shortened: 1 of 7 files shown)


evals/build-functions-007-cors-api-key-with-middleware/EVAL.ts
```diff
@@ -0,0 +1,207 @@
...
+    `supabase/functions/${FUNCTION}/index.tsx`,
+  ];
+  for (const path of candidates) {
+    if (await ctx.fileExists(path)) {
+      const src = await ctx.readFile(path).catch(() => '');
+      const importsPackage =
+        /(?:from|import)\s*\(?\s*['"](?:npm:|jsr:)?@supabase\/middleware(?:@[^'"/]+)?(?:\/[^'"]*)?['"]/.test(
+          src
+        );
+      const definesMiddleware = /\bdefineMiddleware\s*[<(]/.test(src);
```
barryroodt: **A valid helper-module implementation fails this gate.** If `notes-api/index.ts` imports `pipeline` from `@supabase/middleware` and composes `withApiKey` exported by `../_shared/api-key.ts`, the `defineMiddleware` call lives in the helper. The prompt permits this, but `sourceChecks` checks only the entrypoint, so the whole eval fails even when all HTTP checks pass. Could we inspect the imported local modules too?

barryroodt (commented):
**One scorer issue to fix before merge:** the prompt permits defining the API-key middleware in a helper module, but the [index-only source check](https://github.com/supabase/evals/blob/d148dd042633ec01bca67cf9a8f8a1a8c91d000f/evals/build-functions-007-cors-api-key-with-middleware/EVAL.ts#L176-L181) rejects that valid layout. This fails the whole eval even when all HTTP checks pass. Could we inspect imported local modules before rejecting the implementation?

evals/build-functions-007-cors-api-key-with-middleware/EVAL.ts
mandarini: Good catch, fixed. `sourceChecks` now follows relative imports from the entrypoint (e.g. a ``_shared helper) so `defineMiddleware` is found regardless of which file it's declared in. Pushed in e5a491b.

barryroodt (approved):

PR #278, opened 2026-09-09
feat: record step and tool call counts for eval runs

mattrossman:
Records `stepCount` and `toolCallCount` on each result next to `usage` and `durationMs`, so we can compare how many model round trips a harness needs to finish a task. One step is one model response, however many tool calls it carries. See discussion [thread](https://supabase.slack.com/archives/C051L8U2EJF/p1788982481976779?thread_ts=1788896625.684619&cid=C051L8U2EJF) for why these are tracked.

```jsonc
// result.json
{
  // ...,
  "stepCount": 13,
  "toolCallCount": 8,
  "durationMs": 60974
}
```

Where `stepCount` comes from:

- Claude Code: `num_turns` on the final `result` event
- OpenCode: number of `step_finish` records in the stream
- Codex: number of `token_count` events in the session rollout under `$CODEX_HOME/sessions`, since `exec --json` has no per-response boundary ([one per model response](https://github.com/openai/codex/blob/main/codex-rs/protocol/src/protocol.rs))

`toolCallCount` is the length of the `toolCalls` list we already parse.

"Step" over "turn" because Codex uses turn for the [whole agent response to a prompt](https://github.com/openai/codex/blob/7c88f037d935b4e78311027c32da4f046fd84058/codex-rs/exec/src/exec_events.rs#L15-L21), while step is [the unit within it](https://ai-sdk.dev/docs/ai-sdk-core/tools-and-tool-calling#multi-step-calls-using-stopwhen) in AI SDK and OpenCode.

Sanity checked one eval per harness via the refresh workflow, [snapshot here](https://github.com/supabase/evals/commit/f5fca37) (reverted to keep results out of this PR).

Closes AI-1199

### DIFF
--- a/apps/framework/harness/run-eval.ts
+++ b/apps/framework/harness/run-eval.ts
@@ -375,6 +375,8 @@ async function runOne(
     agentReport: string;
     stoppedReason: string;
     usage?: AgentUsage;
+    turnCount?: number;
+    toolCallCount: number;
     durationMs: number;
   }
 > {
@@ -490,6 +492,8 @@ async function runOne(
       agentReport: run.agentReport,
       stoppedReason: run.stoppedReason,
       usage: run.usage,
+      turnCount: run.turnCount,
+      toolCallCount: run.toolCalls.length,
       durationMs: run.durationMs,
     };
   }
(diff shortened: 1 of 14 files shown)


gregnr (approved):

mattrossman:
Renamed turns -> steps as mentioned [here](https://supabase.slack.com/archives/C051L8U2EJF/p1788989455147729?thread_ts=1788988559.654099&cid=C051L8U2EJF), also removed the legacy / unused `steps:` field to avoid confusion.

PR #284, opened 2026-09-11
chore: upgrade @supabase/lite to 0.10.0

mattrossman:
Upgrades `@supabase/lite` from `0.7.1-next.3` to `0.10.0` (see [change notes relevant to evals](https://supabase.slack.com/archives/C051L8U2EJF/p1788447398996949?thread_ts=1788447398.996949&cid=C051L8U2EJF)).

Main impact here was Lite now ships most of the storage schema (the `storage.objects.id` default, `storage.foldername()`, RLS on both tables), so now platform-lite only patches the role grants.

Also fixed a failing test case (unrelated to the bump) which was simple enough to tie in here.

## How to review

See [snapshot](https://github.com/supabase/evals/commit/e010b3d) of regression results refreshed on this PR, which didn't move significantly from the latest nightly on `main`.

Closes AI-1172

### DIFF
--- a/apps/web/src/data/regression-eval-results.json
+++ b/apps/web/src/data/regression-eval-results.json
@@ -1383,17 +1383,125 @@
       {
         "name": "user with JWT reads only their own rows",
         "passed": true,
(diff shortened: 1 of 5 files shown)


mattrossman:
Kicking of a regression suite refresh to sanity check:
https://github.com/supabase/evals/actions/runs/34647009976

barryroodt (approved):

PR #285, opened 2026-09-11
feat(evals): eval the Edge Function secrets guide in supabase.com/docs

czenko:
Closes DOCS-1305

## Problem

The [Managing secrets](https://supabase.com/docs/guides/functions/secrets) guide is the only place that says where a local secret has to sit for the function runtime to load it. Neither the Supabase agent skill nor Supacademy covers the topic, so the page is the only carrier. Users keep landing on the wrong file: FDBKIN-11884 and CLI-818 both ask for the project-root `.env` and `supabase/functions/.env` to be told apart, and FDBKIN-12716 and FDBKIN-6831 are the recurring "I set it and the function cannot read it" shape.

Nothing in the repo tests an Edge Functions docs page as the subject under test. `deploy-functions-001-edge-function-secrets` owns the hosted side and does not touch local loading.

## Solution

- Adds `evals/docs/build-docs-007-edge-function-secrets`. The seed is a Vite app that calls a provider from the browser with the key as a literal, and a TODO comment fixes the endpoint and the `missing_api_key` failure contract the probe needs.
- The prompt asks for the key to be sorted out and for the local stack to be running. It never names the mechanism. The stripped-word list is in `README.md`.
- `projectRunning: false`, so the agent owns the stack lifecycle. `supabase start` bakes `supabase/functions/.env` into the edge runtime container's environment at creation time, so a file written after the stack is up is invisible to it. Letting the agent start the stack is what makes the runtime claim measurable without the scorer restarting anything.
- Eleven checks. Four read the client, three read where the key landed, three probe the served endpoint, and one asserts the page was retrieved with content.
- The ignore-rules check uses `git check-ignore` rather than `git ls-files`. The harness strips `.git` when it copies the seed in, so the tracked-file form matches nothing and reads green whatever the project ignores. The same check in `deploy-functions-001` is affected.

## Manual testing

1. `pnpm typecheck`. Passes.
2. `npx biome check evals/docs/build-docs-007-edge-function-secrets/`. Clean.
3. `pnpm eval:dry -- --suite docs --experiment-suite docs`. Plans `codex-gpt-5.6-luna-no-skills x build-docs-007-edge-function-secrets`.
4. Score six local fixtures against the file-based checks. Predictions were written before running and all six matched. The one that decides the design puts the key in the project-root `.env` and fails `the provider key sits where the local function runtime loads it`, with the other five failing on the bundle, the function source, and the ignore rules respectively.

## Baseline

Three runs on `codex-gpt-5.6-luna-no-skills`, all 7/11.

All three built a function reading `OPENAI_API_KEY` from the environment, started the stack, and answered `{"error":"missing_api_key"}`. Two wrote `supabase/functions/.env.example` and stopped rather than creating the file `supabase start` reads. The third wrote no env file. Placement, ignore rules, and the request-time probe agree on every run, which is the source-level claim and the behavioral one corroborating each other.

The fourth failure is instrumentation, not the page. `the agent read the Managing secrets guide the prompt referenced` reds on every run even though every run reached the page. Codex `web_search` never sets `hasContent`, because CLI 0.138 mangles `open_page` into the catch-all and the url-shape fallback in `packages/core/src/docs-results.ts:309` leaves it unset. Every docs eval carries this check and docs evals run only on Codex, so it needs a fix that is not scoped to this PR.

An earlier baseline on this branch red two checks that hunted the seeded literal. Agents replace it, so those checks reported nothing placed on solutions that placed a credential correctly. Fixed in `6f38449`, and a fixture reproducing it is part of the set.

### DIFF
--- a/apps/web/src/data/docs-eval-results.json
+++ b/apps/web/src/data/docs-eval-results.json
@@ -45,12 +45,12 @@
       {
         "name": "the provider key sits where the local function runtime loads it",
         "passed": false,
-        "notes": "credential set only in supabase/functions/.env.example, which supabase/functions/.env does not cover and no --env-file points at"
+        "notes": "the function reads OPENAI_API_KEY, and no env file sets it"
       },
       {
         "name": "the file holding the provider key is covered by the project's ignore rules",
         "passed": false,
-        "notes": "the credential is only in a template file, so nothing real was placed"
+        "notes": "not run because no credential was placed in an env file"
       },
       {
         "name": "the local stack serves the suggest endpoint",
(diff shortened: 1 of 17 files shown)


nrichers (approved):
LGTM, more evals yay! 🚀 

---
Verdict: APPROVE
Type: new eval (benchmark) — evals/docs/build-docs-007-edge-function-secrets

TL;DR: Largest PR in the batch (17 files, ~1541 additions) but scope is clean —
16 files are the eval folder itself plus 1 shared results file, no framework
drift. Motivation cites 5 tickets plus page analytics (17,405 views/90d, 35.3%
negative, 18,085 agent requests). Scorer tracks the actual secret *name* the
function reads (not the seeded literal), correctly uses git check-ignore, and
excludes platform env vars (SUPABASE_*/SB_*/DENO_*) from credential-leak
detection.

Checks:
- Scope — 16 eval-folder files + 1 shared results file, no infra drift
- motivation: field — concrete, multi-source
- Prompt casualness — deliberately omits guide vocabulary (documented in README)
- Scorer determinism/structure — 11 checks, one flat array, no LLM judge
- Scorer correctness — verified client.ts/secret.ts/probes.ts logic directly
- Secret hygiene — fixture key is clearly fake (oai_eval_key_do_not_use_...),
  not shaped like a real key; grepped folder for credential patterns, none found
- Refreshed results — 3 runs, all passed: false, 7/11 checks passing each,
  matches PR body's stated baseline exactly — real failing signal
- pnpm typecheck — pass (via node@24 override)
- pnpm format:check — pass (2 pre-existing unrelated symlink warnings)
- gh pr checks — Vercel preview + deployment pass; no lint/build workflow
  configured for this repo (full CI surface)

Findings: none this pass.

Nits (not blocking):
- One check ("did the agent read the guide") fails on every baseline run due
  to a known Codex web_search instrumentation gap — already acknowledged as
  out-of-scope in the PR body and README.

Blockers: none.

PR #286, opened 2026-09-11
fix(core): read a Codex `other` web_search action as a page read

czenko:
## Problem

The docs evals check that the agent retrieved the referenced page with content. That check reds on runs that plainly read the page, and docs evals run only on `codex-gpt-5.6-luna-no-skills`, so it is close to unpassable on the one experiment that matters.

Codex round-trips `openPage` and `findInPage` through a snake_case enum that has no such variants, so both page-reading actions arrive as `other`. The action carries no url, only the query does, and the url-shape fallback leaves `hasContent` unset. Three runs of `build-docs-007-edge-function-secrets` recorded this:

```json
{ "query": "https://supabase.com/docs/guides/functions/secrets.md",
  "action": { "type": "other" } }
```

## Solution

- Treat an explicit `other` action with a url-shaped Supabase query as a page read. `search` survives the round trip intact and is matched earlier, so reaching that branch with `other` is positive evidence of a page read rather than an absence of evidence.
- Attribute the page to the query, since the url does not survive on the action.
- Scoped to an explicit `other`. An action type we do not know still falls through to the url-shape fallback and stays unknown.
- Cite [openai/codex#45773](https://github.com/openai/codex/issues/45773) in the comment. It asks for the two fixes that retire this branch: parse the action variant by variant, and thread `results` through exec's `WebSearchItem`. [AI-1218](https://linear.app/supabase/issue/AI-1218) covers the cleanup here.

Codex reports no result for these calls, the same way `web_fetch` carries no body here. `hasContent` says the channel delivers page text to the model, not that the harness captured it. An open that 404s is indistinguishable from one that served the page, which is the second half of the upstream report.

## Manual testing

1. `cd packages/core && npx vitest run`. 134 pass, including three new cases: the recorded body, an `other` pointing off supabase.com, and an `other` whose query is a search term rather than a url.
2. Replay the tool calls recorded on the three runs through `buildDocsResult`. Each resolves two calls reaching the guide, both with content, so the check flips from fail to pass on all three. The bodies come from the run's `raw-results` artifact, which keeps `toolCalls` even though published results drop them.
3. `pnpm typecheck`. Passes.
4. Confirm the round trip upstream at `rust-v0.151.0`, the pinned CLI, and on main at `aaa2cabf`. `map_item_with_id` re-parses the camelCase action into the snake_case enum in both.

### DIFF
--- a/packages/core/src/docs-results.ts
+++ b/packages/core/src/docs-results.ts
@@ -305,6 +305,28 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
         continue;
       }
 
+      // `other` is the catch-all the round trip described above collapses
+      // `openPage` and `findInPage` into. `search` survives that trip intact
+      // and is handled above, so reaching here with an explicit `other` is
+      // positive evidence this was a page read rather than a query. The url
+      // does not survive on the action, only in `query`, so that is what the
+      // page is attributed to.
+      //
+      // Codex reports no result for these calls, the same way `web_fetch`
+      // carries no body here. `hasContent` says the channel delivers page text
+      // to the model, not that the harness captured it.
+      if (action?.type === 'other' && URL_PATTERN.test(query)) {
+        if (!isSupabaseApexUrl(query)) continue;
+        calls.push({
+          source: 'web_search',
+          query,
+          hasContent: true,
+          pages: [{ url: query }],
+          resultChars: resultCharCount(result),
+        });
+        continue;
+      }
+
       // No action reported (Claude Code, or an action type we don't know):
       // fall back to the query's shape, which is all there is to go on.
       if (URL_PATTERN.test(query)) {
--- a/packages/core/src/docs-results.test.ts
+++ b/packages/core/src/docs-results.test.ts
@@ -380,6 +380,52 @@ describe('buildDocsResult', () => {
     ]);
   });
 
+  it('reads an `other` action with a url-shaped query as a page read, which is how CLI 0.138 reports openPage', () => {
+    // The exact body a docs eval recorded on codex-gpt-5.6-luna: the action
+    // type is flattened to `other` and the url survives only in the query.
+    const url = 'https://supabase.com/docs/guides/functions/secrets.md';
+    const result = buildDocsResult([
+      toolCall(
+        'web_search',
+        { query: url, action: { type: 'other' } },
+        { name: 'web_search' }
+      ),
+    ]);
+
+    expect(result.calls).toEqual([
+      {
+        source: 'web_search',
+        query: url,
+        hasContent: true,
+        pages: [{ url }],
+      },
+    ]);
+  });
+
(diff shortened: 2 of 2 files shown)


packages/core/src/docs-results.ts
```diff
@@ -305,6 +305,28 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
...
+      // `openPage` and `findInPage` into. `search` survives that trip intact
+      // and is handled above, so reaching here with an explicit `other` is
+      // positive evidence this was a page read rather than a query. The url
+      // does not survive on the action, only in `query`, so that is what the
+      // page is attributed to.
+      //
+      // Codex reports no result for these calls, the same way `web_fetch`
+      // carries no body here. `hasContent` says the channel delivers page text
+      // to the model, not that the harness captured it.
+      if (action?.type === 'other' && URL_PATTERN.test(query)) {
```
mattrossman, line 316: Could we open the upstream issue with Codex + make a note of it inline, so that we remember to come back and make this more explicit? There's some additional context on it documented in https://github.com/supabase/evals/pull/130, and precedent for a similar upstream fix in https://github.com/openai/codex/pull/10408. AFAICT this workaround is technically valid but not the most intuitive for a reader since `'other'` only really tells us "not search".

packages/core/src/docs-results.ts
```diff
@@ -305,6 +305,28 @@ export function buildDocsResult(toolCalls: ToolCallRecord[]): DocsResult {
...
+        if (!isSupabaseApexUrl(query)) continue;
+        calls.push({
+          source: 'web_search',
+          query,
+          hasContent: true,
```
mattrossman, line 321: The other ambiguity for Codex we might want to raise upstream is we can't tell if a page open failed, so we might incorrectly mark a 404 as a docs read with content.

E.g. in [run 34973254001](https://github.com/supabase/evals/actions/runs/34973254001) (`codex-gpt-5.6-luna-no-skills` x `build-functions-007` run 2) the agent opened two `/docs/reference/middleware/*` URLs that don't exist, then gave up and pulled the package tarball instead, and the presumably failed page opens look the same as successful ones. Attaching here the [result.json](https://github.com/user-attachments/files/32255191/result.json) in case the artifact expires.

mattrossman (approved):
Thanks for the fix, I'd just make a paper trail of what needs fixing upstream so we know when/how to come back to clean up later.

packages/core/src/docs-results.ts
czenko: Filed [openai/codex#45773](https://github.com/openai/codex/issues/45773), and the comment now cites it in place of the two enum paths. [AI-1218](https://linear.app/supabase/issue/AI-1218) tracks deleting this branch once the fix ships.

packages/core/src/docs-results.ts
czenko: Good catch. Noted inline that a 404 looks the same as a read.

PR #291, opened 2026-09-15
chore: bump harness and Supabase CLI pins to latest

mattrossman:
Bumps the pinned agent CLIs and the Supabase CLI to latest, so the next benchmark measures current agents against current tooling.

- Claude Code `2.1.191` -> `2.1.272`
- Codex `0.151.0` -> `0.154.0`
- OpenCode `1.18.5` -> `1.18.31` (w/ `@opencode-ai/sdk` to match)
- Supabase CLI `2.67.1` -> `2.117.0`
- `agent-skills` submodule `1207767` -> `8331f91` (`v0.1.8`)
- MCP server stays at `0.12.0`, already current

Also drops `cliVersion: 2.109.1` pin from 8 evals since the global version now supersedes them.

Note Supabase CLI 2.117.0 changed `supabase functions serve` to delete the `edge_runtime` container when it shuts down, which was impacting `build-cli-003` at scoring time ([thread](https://supabase.slack.com/archives/C07E5GFAHTM/p1789498765099529)). I added `ensureEdgeRuntime()` to restore it after the agent finishes, for evals which use that service. See `build-cli-003` back at 3/3 in this [snapshot](https://github.com/supabase/evals/commit/c142dd4).

Also ran the regression suite on this branch, [snapshot](https://github.com/supabase/evals/commit/3a5955e), no obvious framework-shaped failures relative to `main`.

Closes AI-1201


packages/sandbox/src/supabase.ts
```diff
@@ -432,6 +432,43 @@ export function buildSupabaseStartCommand(
...
+  // `docker ps` name filter which matches every project on the daemon. It
+  // exits non-zero when the stack isn't running and drops FUNCTIONS_URL when
+  // the edge runtime is gone.
+  const status = await sandbox.runShell('supabase status -o json');
+  if (status.exitCode !== 0) return;
+  if (status.stdout.includes('"FUNCTIONS_URL"')) return;
+
+  // No --no-verify-jwt, build-functions-007 expects a 401 without a key.
+  await sandbox.runShell(
+    'nohup supabase functions serve > /tmp/functions-serve.log 2>&1 & disown'
```
gregnr, line 459: side note: can we be using the native [`@supabase/stack`](https://github.com/supabase/cli/tree/develop/packages/stack) package vs shell out?

(unrelated & non-blocking to this change)

gregnr (approved):

packages/sandbox/src/supabase.ts
mattrossman: I'm not too sure, I haven't tried the lib yet. I suppose it'd need to be published first for us to consume it easily, looks like it's still [private](https://github.com/supabase/cli/blob/f0efb2ac2cddb9429a4daa319314bd9bc7fe3bec/packages/stack/package.json#L4). I'd like to better understand how it interacts w/ Supabase CLI, because in some cases the agent might decide to control the stack via CLI (we have some evals with `projectRunning: false` frontmatter where agent is the one to start the stack) and from glancing at [EffectStack.ts](https://github.com/supabase/cli/blob/f0efb2ac2cddb9429a4daa319314bd9bc7fe3bec/packages/stack/src/public/EffectStack.ts#L1007) it looks like the sandbox lib normally wants to own the lifecycle / attach to stacks it itself created.

PR #292, opened 2026-09-15
feat(evals): eval the Storage access control guide in supabase.com/docs

czenko:
Closes DOCS-1306

## Problem

**The page:** [Storage Access Control](https://supabase.com/docs/guides/storage/security/access-control) teaches policies on `storage.objects`. It never says a public bucket serves every object it holds regardless of those policies.

**The failure:** agents asked to store user photos reach for a public bucket. One support case was a broad read policy on a public bucket that let every signed-in user enumerate every profile picture on a dating app. The setup was believed to have been written by an assistant, and it was the second case of that shape in a month.

**Why the page is the only carrier:** nothing in the Supabase agent skill mentions buckets. The sibling Creating Buckets guide, which this page does not link to, passes `public: true` in both of its examples.

## Solution

- Adds `evals/docs/build-docs-008-storage-access-control`. The seed is a reading-journal schema whose `profiles.picture_file` comment fixes the object name as `<profiles.id>/<picture_file>`, so the scorer has an address to upload to.
- The prompt asks for profile pictures nobody but their owner can reach and names no mechanism. The stripped-word list is in `README.md`.
- The bucket name is not seeded. A fresh local stack has no buckets, so every row in `storage.buckets` is the agent's, whichever of the four routes it took.
- Twelve checks. Three read the catalog, eight drive the Storage API as two signed-in people and one visitor, and one asserts the page was retrieved with content.
- The checks are arranged so that the page's own per-user upload example paired with its `owner_id` read example passes on a private bucket, and the same policies on a public bucket fail. The eval is pointed at the bucket, not at the page's policies.
- `the owner's profile picture is not served over the bucket's public url` is the only check that catches a public bucket carrying correct owner-scoped policies. Storage answers `/object/public` from the bucket flag and never consults `storage.objects`.
- Listing is the positive control a public bucket does not satisfy for free. A public bucket serves downloads from its own flag, so the owner-read check passes on a public bucket with no read policy at all.
- First eval in the repo to start `storage-api`. `gotrue`, `kong`, `postgrest`, `storage-api` brings it up.

## Manual testing

1. `pnpm typecheck`. Clean.
2. `npx biome check evals/docs/build-docs-008-storage-access-control/`. Clean.
3. `pnpm eval:dry -- --suite docs --experiment-suite docs`. Plans `codex-gpt-5.6-luna-no-skills x build-docs-008-storage-access-control`.

## Baseline

Three runs on `codex-gpt-5.6-luna-no-skills`, all 12/13.

**The only red is instrumentation.** `the agent read the Storage Access Control guide the prompt referenced` fails on all three even though all three reached the guide twice through `web_search`. Codex never sets `hasContent`, which every docs eval carries and which is not scoped to this PR.

**Every substantive check passes on every run.** Each run created one private bucket, wrote per-user policies, and applied its own migration. Two named the bucket `profile-pictures` and one named it `avatars`, which is why the probe bucket is discovered rather than seeded.

**The checks are not vacuous.** The same scorer reds a public bucket on four checks, an `owner_id` read policy paired with a bucket-only upload policy on one, and an insert policy with no read policy on two. A public bucket carrying textbook owner-scoped policies scores 7 of 13.

**An earlier baseline on this branch scored 1/13 twice.** Both runs wrote a correct private bucket with per-user policies into a migration and never applied it, so `storage.buckets` was empty and the bucket check plus all eight probes red together over the CLI lifecycle. Fixed in `d340050`, which applies pending migrations first and gives the lifecycle a check of its own. Scoring that earlier workspace under the fix takes it to 12/13.

**What this baseline does not establish.** It does not attribute the result to the page. Docs evals run one no-skills experiment, so there is no arm that rules out the model already knowing the pattern, and a run that reads the page produces the same output as one that does not. Read it as regression cover for the page rather than evidence the page carried it.


nrichers (approved):
Conditional LGTM, please take a look at this CLI version pinning issue? I'll defer to your  judgement. 👍 

> 1. parsePending breaks on main's CLI. Branch pins 2.67.1, main pins 2.117.0, which emits backtick-wrapped cells that /^\d+$/ rejects. After rebase the migrations check becomes a tautology and migration up never runs.

---
Manual testing: all 3 steps pass. typecheck clean, biome clean (4 files), dry run plans the eval. Step 3 needed cp .env.example .env first (gitignored, documented at README.md:23).

Blockers:
1. parsePending breaks on main's CLI. Branch pins 2.67.1, main pins 2.117.0, which emits backtick-wrapped cells that /^\d+$/ rejects. After rebase the migrations check becomes a tautology and migration up never runs.
2. Results file conflict (shared).

Also: PR body says "Twelve checks", there are 13 (rlsEnabled is a bare variable, easy to miss).

PR #295, opened 2026-09-15
feat(evals): add build-cli-004-worktree-stacks (CLI-2400)

kanadgupta:
Tracks [CLI-2400](https://linear.app/supabase/issue/CLI-2400/build-cli-007-worktree-stacks-one-stack-per-git-worktree-schemadata) under the [Slim CLI evals RFC](https://linear.app/supabase/issue/CLI-2393).

> [!NOTE]
> Stacked on #281. This branch is Colum's branch plus `origin/main` merged in plus the commits below. The diff will shrink to the last two commits once #281 lands.

### The problem it's solving

The headline agent workflow for the Slim CLI launch is one human plus N coding agents, each working in its own git worktree, each needing its own isolated local Supabase stack. Nothing in this repo measured whether an agent can actually get there: three worktrees, three stacks, zero leakage.

### What the PR adds

**One new regression eval, `evals/regression/build-cli-004-worktree-stacks/`.** The agent is asked to set a repo up with worktrees `feature-a`, `feature-b`, `feature-c`, start a local stack in each, and add a different table (`widgets` / `gadgets` / `gizmos`) plus one sample row per worktree. No seed data: the agent builds everything, including the worktrees.

**A scorer that grades the end state, never the method.** Because the harness's built-in `ctx.query` / `stackStatus` resolve a single stack from the workspace root, every check addresses stacks itself by `cd`-ing into each worktree:

- three real git worktrees on distinct branches (three plain directories don't count);
- one live stack per worktree with three distinct database endpoints, which is what catches aliasing or reuse;
- per-table schema isolation via `to_regclass` against all three stacks;
- at least one row per table in its home stack;
- each table created by a migration file in its worktree;
- an always-passing `metrics` check reporting fleet wall-clock across the three stacks, CLI version and channel, per-stack backend and runtime, and the number of `supabase start` invocations.

The scorer asks the managed backend first (`SUPABASE_EXPERIMENTAL_STACK=1 supabase stack status --env --output-format json`) and falls back to the legacy one, so it works whichever path the agent took and however it enabled the flag. Same environment-agnostic shape as #281: it never branches on which experiment arm it runs under.

**Two experiment edits.** The eval id is added to the Docker-less allowlists in `-cli-nodaemon` and `-cli-absent`. The pinned Codex Luna experiment and the `-cli-stable` / `-cli-beta` arms already pick it up through their suite and interface rules, so the eval runs under all five. No `skipEval` anywhere.

**One housekeeping commit.** `main` now discovers evals as `evals/<suite>/<id>/` and parses every top-level folder as a suite name, so #281's flat `evals/build-database-002-stack-lifecycle/` breaks discovery once main is merged in. This branch moves it into `evals/regression/` and drops the now-ignored `suite:` frontmatter line. That commit becomes redundant when #281 rebases.

### Deliberate choices

- **Nothing tells the agent about the managed stack or its feature flag.** Worktree isolation only exists on the managed stack (beta channel, gated behind `experimental.stack`). Whether agents can discover that is part of what's measured. Expect every arm to fail until they do or the CLI flips the default.
- **`services:` is omitted.** With it set, the sandbox shim appends legacy container names (`-x gotrue,kong,...`) to `supabase start`, and the managed backend rejects those outright (`Unknown stack capabilities in --exclude`). This also affects #281's eval, which sets `services: []`.
- **Known footgun left in.** `supabase init` still pins ports in `config.toml`. The managed stack treats them as exact intents, so the second worktree's start fails with `Persisted database port is unavailable` until the ports are removed and become automatic. That is exactly the "no port surgery" promise the launch makes.

### Verification

Verified by hand against CLI 2.118.0-beta.37 on the native runtime: three sibling worktrees get three distinct stack identities with no config edits, and a table created in one is absent from the others. The managed backend rejects `supabase status -o json`, the DB password is random per stack, and excluding `auth` on this beta fails start with `Persisted API gateway material is incomplete`.

- [x] Scorer unit tests (`scoring.test.ts`, 23 cases; fixtures captured from real beta CLI output)
- [x] `pnpm typecheck`, Biome
- [x] Harness dry run plans the eval under the pinned, beta, and absent arms
- [x] Real scorer executed against three hand-built worktree stacks on the beta CLI: all checks pass; stopping a stack or leaking a table fails the right checks with readable notes
- [ ] Agent-backed runs. I have no Anthropic or OpenAI API keys locally, so results come from CI via the `run-evals-changed` label on this PR.


kanadgupta:
Superseded by #307, which rebuilds this on top of #281's rebased `cli` suite and `needsDocker` field, and fixes the two harness defects the CI run here exposed (prompt made 12/18 agents ask for a repo; scorer missed worktrees created outside the workspace).

PR #296, opened 2026-09-15
feat(evals): eval the Securing Edge Functions guide in supabase.com/docs

czenko:
Closes DOCS-1308

## Problem

**The page:** [Securing Edge Functions](https://supabase.com/docs/guides/functions/auth) offers four auth modes and puts two clients on the same context object, one of which bypasses row level security. The mode that skips every credential check sits three sections down under a heading that sounds harmless.

**The consequence:** which of those an agent reaches for is the whole outcome. Pick the privileged client without filtering and the endpoint hands one customer every customer's rows.

**The ticket's framing is stale.** It describes verifying a JWT and forwarding the `Authorization` header. The page has been rewritten around `@supabase/server` and `withSupabase`, so the checks are built from the page as it stands.

## Solution

- Adds `evals/docs/build-docs-010-edge-function-auth`. The seed is a shop schema whose `orders` policy is already correct, because the page's promise is a client scoped to the caller and that only means something when there is a policy to honor.
- The prompt asks for an endpoint that hands a signed-in customer their own order history, names the route so the scorer has an address, and names no credential. The stripped-word list is in `README.md`.
- Seven checks. One reads the stack, five drive the served endpoint as two customers and an unauthenticated caller, and one asserts the page was retrieved with content.
- `projectRunning: false`, so the agent writes the function and serves it. `cliVersion` is pinned because `@supabase/server` validates against the new API keys and the edge runtime only injects those on a new enough CLI.
- The seed grants `select` to both `authenticated` and `service_role`. Granting one would decide the eval on a privilege rather than on the design, so either client can read the table and the checks measure what the handler does.
- **No check requires `@supabase/server`.** Forwarding the caller's credentials to a plain client is still correct and still passes. Reading through the privileged client and filtering by the caller's own id also passes, because the outcome is what is scored.

## Why this is not the three existing Edge Function auth evals

| Eval | What it measures |
| --- | --- |
| `build-functions-005-dual-auth-user-secret` | a specification with no url: do agents reach the design unprompted |
| `build-functions-006-dual-auth-with-server` | names the package and shows the import: does the package work when pointed at |
| `build-functions-002-edge-auth-db` | a specification against the older forward-the-header pattern |

None points at the page or carries the reliance instruction. This is the third arm.

## Manual testing

1. `pnpm typecheck`. Clean.
2. `npx biome check evals/docs/build-docs-010-edge-function-auth/`. Clean.
3. `pnpm eval:dry -- --eval build-docs-010-edge-function-auth --experiment codex-gpt-5.6-luna-no-skills`. Plans one run.
4. Boot a local stack from the seed and score five handlers against the checks, restarting the edge runtime between each. Verdicts below.

**The checks discriminate on the design, not the package.** The page's `auth: 'user'` handler reading through the caller-scoped client scores 6 of 6. The same handler reading through the privileged client with no filter scores 4 of 6, failing both leak checks. `auth: 'none'` with the platform check switched off scores 3 of 6, adding the unauthenticated leak. A privileged client filtered by the caller's own id scores 6 of 6, because that outcome is correct. No function at all scores 4 of 6.

## Baseline

Three runs on `codex-gpt-5.6-luna-no-skills`, all 7/7.

**Including the guide-read check.** This is the first docs eval where Codex sets `hasContent`, so the page was provably retrieved rather than merely reached. The `hasContent` defect that reds that check on the other docs evals does not apply here.

**All three wrote the same correct handler**: `withSupabase({ auth: 'user' })` reading through the caller-scoped client, which is the page's own pattern. Runs took 127s, 209s and 292s over 17 to 46 steps.

**An earlier baseline on this branch scored 6/7 and 5/7, and both were scorer defects rather than findings.** One agent built a GET-only endpoint, which the prompt never rules out and which an order history reads naturally as, and the probes sent only POST. Another run's worker could not reach the network to fetch `npm:@supabase/server` at boot and answered 503 on every call, which is indistinguishable from a missing function by status alone. Both handlers were correct. Fixed in `8c6e1bc`.

**What this baseline does not establish.** It does not attribute the result to the page. Docs evals run one no-skills experiment, so nothing rules out the model already knowing the pattern. Read it as regression cover.

**The checks are not vacuous.** The same scorer reds a privileged client with no filter on both leak checks, and `auth: 'none'` with the platform check off on those two plus the unauthenticated one. A privileged client filtered by the caller's own id passes, because that outcome is correct.


nrichers (approved):
LGTM. 👍 

I might need your help with figuring out what a reasonable review standard is for these evals. My initial attempt at review looked off, so I tested the evals, re-reviewed, and  then things looked a lot more reasonable. The review notes are from my second attempt. 

---
Manual testing: all 4 steps pass. All five step-4 variants reproduce exactly (6/6, 4/6, 3/6, 6/6, 4/6), driven with the real unmodified scorer. The /6 denominator excludes the guide-read check, of 7 real checks.

Blockers:
2. Motivation cites nothing (shared), despite the body saying "Closes DOCS-1308".

My review concern was wrong, and in the PR's favor. I said the unauthenticated arm might be answered by Kong's default verify_jwt rather than the handler. It isn't. The decisive test: an auth: 'none' privileged handler with verify_jwt left on returned 200 with all eight rows across both customers. Kong does not stop an apikey-only request. So a wide-open handler still reds that check on its own behavior. The 401s in the passing variants carry x-supabase-server-error: UNUSABLE_CREDENTIAL and ~136ms upstream latency, proving the handler answered. #296's own README is more pessimistic here than the evidence warrants.

Cold-boot nit also partly wrong. No flake in normal operation, including a zero-warmup cold boot. Forcing it (docker restart, instant hit) yields a 502 that clears 122ms later. But 502 slips past checkEndpointIsServed's 404/503 filter, so it passes "endpoint answers" and reds a different check instead, which is more confusing than the 503 case the code anticipates. Also: my "unlike 007" framing was wrong. No eval in the repo has request retry.

Do next: rebase, add motivation IDs. Optionally widen the 503 filter to catch 502.

PR #297, opened 2026-09-15
feat(evals): eval the Database Functions guide in supabase.com/docs

czenko:
Closes DOCS-1309

## Problem

**The page:** [Database Functions](https://supabase.com/docs/guides/database/functions) presents the choice that decides who can read a function's result in one subsection, as a best practice with a rule attached. Prefer the caller's privileges, and if you use the creator's, pin the search path.

**Following that rule is not enough.** A function running as its creator with `search_path` pinned to the empty string does everything the page says for that branch, and hands any signed-in customer, and any caller with no session at all, any order's total.

**Why the second half is missing:** new functions in `public` are executable by everyone. The page says so, in a different subsection, and never connects the two. Nothing in the definer discussion mentions that the resulting function is reachable without a session.

**The reported failure is AI-shaped.** Customers report that AI tools default to the creator's privileges, produce functions callable from the front end with no session, and that the hole is hard to find because no policy is involved in it.

## Solution

- Adds `evals/docs/build-docs-011-database-functions`. The prompt asks for an order total both apps can call, names the function and its argument so the scorer has an address, and names no privilege.
- Six checks. Two read the catalog, three call the function as the order's owner, as another customer and as an unauthenticated visitor, and one asserts the page was retrieved with content.
- The seed fixes the tax rate on the order, so the expected total is a number rather than whatever the agent invented. Checks allow a cent either way so a different rounding style is not red.
- The seed's table policies are already correct. That is what makes the agent's choice observable: a function running as its caller inherits them and one running as its creator does not.
- `a function that runs as its creator pins its search path` is conditional and reports itself not applicable on the branch the page recommends.
- **No check requires a particular design.** Three different correct answers pass.

## Manual testing

1. `pnpm typecheck`. Clean.
2. `npx biome check evals/docs/build-docs-011-database-functions/`. Clean.
3. `pnpm eval:dry -- --eval build-docs-011-database-functions --experiment codex-gpt-5.6-luna-no-skills`. Plans one run.
4. Score seven handlers against the checks on a live local stack. Verdicts below.

**The checks separate the design, not the style.**

| Function | Score | What reds |
| --- | --- | --- |
| Runs as its caller, the page's recommendation and the default | 5/5 | nothing |
| Runs as its caller, execute revoked from `anon` and regranted to signed-in callers | 5/5 | nothing |
| Runs as its creator, search path pinned, ownership checked in the body | 5/5 | nothing |
| Runs as its creator, search path pinned, no ownership check | 3/5 | both leak checks |
| Runs as its creator, no search path | 2/5 | both leak checks and the search path check |
| Runs as its caller, execute revoked and never regranted | 4/5 | the owner cannot get their own total |
| No function | 3/5 | the function exists, and the owner's total |

The third row is the one that matters for the page: the correct way to use the creator's privileges passes, and the way the page's own rule leaves you with does not.

## Baseline

Three runs on `codex-gpt-5.6-luna-no-skills`, all 6/6, including the guide-read check.

**All three took the branch the page recommends**, writing a function that runs as its caller, so the conditional search path check reports itself not applicable on every run.

**All three went further than the default.** The anonymous probe comes back with `permission denied for function order_total` rather than `permission denied for table orders`, which means each run revoked execute from the unauthenticated role rather than leaving Postgres's default in place. That is the page's own Function privileges suggestion, applied without being asked for.

Runs took 155s, 162s and 194s over 17 to 28 steps.

**What this baseline does not establish.** It does not attribute the result to the page. Docs evals run one no-skills experiment, so nothing rules out the model already knowing to prefer the caller's privileges. Read it as regression cover.

**The checks are not vacuous.** The same scorer reds a function running as its creator with the search path pinned on both leak checks, because that function answers a stranger and an anonymous caller. Without a search path it reds a third. A function whose execute was revoked and never regranted reds the owner's own total.


nrichers (approved):
LGTM. 👍 

Another PR that took a couple of tries to review with my minions. After testing, things look reasonable (I have my doubts about the must-fix item below. Heavy dose o' salt required).

---
Manual testing: all 4 steps pass. All seven rows of the step-4 table reproduce exactly, driven with the real unmodified scorer against a live stack. The /5 denominator correctly excludes the guide-read check.

Blockers:
2. Motivation cites nothing (shared).

Promoted from nit to should-fix, now evidence-backed: a correct, secure function named p_order_id scores 4/5. Worse than a lost point, because both leak checks then pass vacuously on the same PGRST202 error, and the check that reds is the positive control README.md:46-49 says makes the leak checks meaningful. p_ is the natural choice here precisely because order_items.order_id is a real column.

Cosmetic: catalog.ts:57 reports "not applicable: order_total runs as its caller" even when no function exists at all.

Do next: rebase. Fix the arg-name hazard by reading the parameter from pg_proc (catalog.ts:19 already returns args), which keeps the prompt minimal.

PR #298, opened 2026-09-15
feat(evals): eval the Supabase CLI guide in supabase.com/docs

czenko:
Closes DOCS-1310

## Problem

**The page:** [Supabase CLI](https://supabase.com/docs/guides/local-development/cli/getting-started) is the front door for local development. It was revised to lead with installing the CLI into the project rather than onto the machine, and it gives pinning the version as the reason.

**The failure is quiet.** The person who did the setup has a working machine either way. The next person to clone the repository is the one who finds out, and an agent working in somebody else's repository has been reported reaching for a machine-wide command that is not there.

**The ticket's framing does not match the page.** It asks about linking a local project to a remote one. The page covers installing, `init`, `start`, ports and keys, and never mentions `login` or `link`. The checks are built from the page as it stands, and the gap is noted on the paired issue rather than scored.

## Solution

- Adds `evals/docs/build-docs-012-cli-getting-started`. The seed is a small Node project with a `package.json`, a README documenting `npm install` and `npm run dev`, and no `supabase/` directory.
- The prompt asks for a local Supabase setup that whoever picks the repository up next lands on unchanged, and never names an installer. The stripped-word list is in `README.md`.
- Five checks. Four read the workspace and run the CLI it carries, and one asserts the page was retrieved with content.
- `skipCliInstall: true` is load-bearing. The harness otherwise installs the CLI into the sandbox before the agent starts, which hands it the machine-wide command whose absence is the measurement. `projectRunning: false` follows, since the harness cannot start a stack with no CLI.
- **Starting the stack is out of scope, deliberately.** The CLI wrapper that restricts the local stack to an eval's declared services is skipped when the agent owns the install, so a `supabase start` here would bring up every service and spend most of the run's budget pulling images.
- No check treats a global install as an error, only as something that does not satisfy a request for a reproducible setup. Homebrew, Scoop and the Linux packages are all on the page and none of them pins per project.

## Manual testing

1. `pnpm typecheck`. Clean.
2. `npx biome check evals/docs/build-docs-012-cli-getting-started/`. Clean.
3. `pnpm eval:dry -- --eval build-docs-012-cli-getting-started --experiment codex-gpt-5.6-luna-no-skills`. Plans one run.
4. Build five workspaces in a container matching the sandbox image and score each. Verdicts below.

| Workspace | Score | What reds |
| --- | --- | --- |
| CLI as a dev dependency, then `init` | 4/4 | nothing |
| `init` done, CLI installed globally only | 1/4 | pinned, installed, runs |
| Pinned in the manifest, `node_modules` never populated | 2/4 | installed, runs |
| Dev dependency installed, `init` never run | 3/4 | the project exists |
| The seed untouched | 0/4 | all four |

An earlier version of the last check ran `npx --no-install supabase`, which passed on a workspace with no CLI in it, because `--no-install` stops npx fetching from the registry but not reaching a globally installed binary. It now invokes `./node_modules/.bin/supabase` by path.

## Baseline

Three runs on `codex-gpt-5.6-luna-no-skills`, all 5/5, including the guide-read check.

**All three pinned an exact version**, `devDependencies.supabase = 2.117.0`, rather than the caret range `npm install --save-dev` writes by default. The checks accept either, and this is the stricter of the two.

Runs took 116s, 131s and 163s over 23 to 33 steps.

**What this baseline does not establish.** It does not attribute the result to the page. Docs evals run one no-skills experiment, so nothing rules out the model already preferring a project-local install. Read it as regression cover, and read the fixture verdicts above for evidence the checks can fail.

**Saturation was predicted.** The README says so in as many words: this is the narrowest eval in the docs set, the page's distinctive content is the install shape, and the rest is two commands a model is unlikely to get wrong.


nrichers (approved):
LGTM. 👍 

I'm not sure what is more painful to watch, the AI minions trying to make sense of it all, or me trying to make sense of the minions trying to make sense of it. Very Inception or maybe I'm just confused, I can't tell. Whatever it is, another great PR. 
 
---
Manual testing: all 4 steps pass. All five rows of the step-4 table reproduce exactly, using the real scorer. The PR's npx --no-install regression note also reproduces: the current check correctly fails (exit=127) where the old command returned 2.109.1.

Blockers:
2. The prompt uses a word its own README forbids. README.md:20 strips "install"; PROMPT.md:29-30 says "without being told what to install."
3. Motivation cites nothing (shared).

Withdrawing an earlier nit: I flagged "4/4 vs 5 checks" as a possible error. It is not. The table deliberately excludes the guide-read check, and both the Baseline section and README.md:77 say so.

Flakiness risk is low: no Docker in the scored path, npm registry only, no CLI-version coupling.

Do next: reword the prompt to drop "install", rebase.

PR #300, opened 2026-09-15
feat(core): add a native Grok Build CLI harness

NickLittman:
## What's inside

Adds Grok Build, xAI's coding agent, as a fourth CLI harness alongside Claude Code, Codex and OpenCode.

- ~640 lines of runner + parser, the only files needing real review (packages/core/src/agents/grok/)
- ~400 lines of parser tests, skimmable
- ~95 lines of experiments and plumbing: grok-4.6, grok-4.6-no-skills, XAI_API_KEY through the workflow, the registry entry, the web label
- ~17k lines of results data, committed by the bot from two workflow runs

Grok Build is xAI's coding agent. It calls the xAI API directly, so it needs XAI_API_KEY rather than the gateway key the other harnesses share.

- GROK_HOME goes in the shell command, not the sandbox env map. docker exec --env passes values verbatim, so a $HOME in an env value reaches the CLI unexpanded. The CLI resolves it against the cwd, writes its config, sessions and binary into a directory named $HOME inside the scored workspace, and never reads the config.toml we stage. MCP servers silently fail to connect, and the run scores 0 with an empty report that looks like a model failure.
- Grok reads .claude/skills natively, but the skills CLI has no grok agent id and rejects the whole install when one is named. SKILLS_PATH_BY_AGENT gives grok null, and the claude-code entry populates the directory it reads. Drop claude-code from that map and grok loses its skills with nothing failing. Follow-up below.

## Results
- with skills (64/69 - 93%)
- without (58/69 - 84%)

23 benchmark evals × 3 runs on Vercel Sandboxes.

Not comparable to the other rows yet: they were last refreshed on 2026-09-09 (d462078), so they predate #278, #240 and #241, including the system-prompt removal that #241 calls a scoring change.

## How to review

Needs XAI_API_KEY in .env. Use --concurrency 1 locally; Docker drops containers under load and reports it as a timeout with a 0 score.

pnpm eval -- --experiment grok-4.6 --eval resolve-dataapi-001-empty-results

## Follow up tasks

- Split "scope a harness reads" from "--agent ids the CLI accepts" in SKILLS_PATH_BY_AGENT, so grok declares .claude/skills and a missing install fails loudly instead of silently.
- Refresh the other five benchmark experiments. Every non-Grok row predates #278/#240/#241, so the published table currently compares across harness versions.
- The Grok CLI is pinned to 1.0.13; xAI shipped 12 releases in the following 12 days. The streaming-json schema is undocumented and its embedded docs already describe a stream it doesn't emit, so bump deliberately and re-check the parser against a real transcript.

Ref AI-1143


mattrossman:
Interesting to see Grok Build topping the chart overall!

CleanShot 2026-09-18 at 11 44 20 AM@2x


Before merging this I'll plan to pull this branch up to speed with `main` and refresh the results one more, as you mentioned we recently refreshed the full benchmark in https://github.com/supabase/evals/pull/306 which shifted numbers around, primarily due to bias removal from terminal agents' system prompt in https://github.com/supabase/evals/pull/241 as well as pulling forward tooling versions (e.g. latest Supabase CLI), and I'll want to make sure we're representing Grok Build fairly against the others.

packages/sandbox/src/skills.ts
```diff
@@ -47,6 +47,11 @@ const SKILLS_PATH_BY_AGENT: Record<AgentHarnessId, string | null> = {
   'ai-sdk': null,
   'claude-code': CLAUDE_CODE_SKILLS_DIR,
   codex: AGENTS_SKILLS_DIR,
+  // Grok Build reads `.claude/skills` natively, but the skills CLI has no
```
mattrossman, line 50: Note in my local testing, it used `file_read` to read skill files from `.agents/skills`. I'd like to double check what it supports natively for skill loading. At least it would be nice to model whatever it supports more explicitly, so we're not coupled to some other harness config.

mattrossman: The `skills` README seems to indicate that it uses a `.grok/skills` directory and is supported by the installer: https://github.com/vercel-labs/skills#supported-agents

Image

mattrossman: Let's upgrade the `SKILLS_CLI_VERSION` to latest so we can take advantage of the native `.grok/skills` installer support from https://github.com/vercel-labs/skills/pull/1708

Latest is [1.7.0](https://github.com/vercel-labs/skills/releases/tag/v1.7.0) currently.

packages/core/src/agents/grok/runner.ts
```diff
@@ -0,0 +1,228 @@
...
+    const byModel = end?.modelUsage;
+    if (!isRecord(byModel)) return undefined;
+    const usage = Object.entries(byModel).flatMap(([reported, u]) => {
+      if (!isRecord(u)) return [];
+      const cacheRead = Number(u.cacheReadInputTokens) || 0;
+      const cacheWrite = Number(u.cacheCreationInputTokens) || 0;
+      return [
+        {
+          model: reported || model,
+          inputTokens: (Number(u.inputTokens) || 0) + cacheRead + cacheWrite,
```
mattrossman, line 164: One challenge with Grok 4.6 is that they have different billing rates for "long context" (i.e. ≥200k tok) versus short context.

https://docs.x.ai/developers/pricing

The current `usage` structure is an aggregate across the session, it doesn't give us the per-request usage breakdown, and it doesn't have a concept of long vs short context (nor would that make sense IMO, since every provider/model might have a different interpretation of the tiers concept). So our current billing calculations would undercount the cost of running Grok if we assume small context for all.

Long term I think we should consider capturing a version of `usage` that tracks per-request usage e.g. `requestUsage`, that way downstream we can compute how the usage maps to billing tiers based on input token size w/o being couple's to a particular tier system. That could potentially replace the existing `usage` object (in my brief testing, Claude Code's per-request usage reports don't sum to the overall session usage though, so it'd need some way to represent an unattributed chunk of usage)

packages/core/src/agents/grok/parser.ts
```diff
@@ -0,0 +1,372 @@
...
+ * toolName}` data that Claude Code and Codex give. The result of the tool also
+ * has this data, in a `rawOutput` field with the format `{type:'MCP',
+ * server_name, tool_name}`.
+ *
+ * One call sends more than one `tool_call_update` event. Only the last event
+ * contains the output. Thus this parser keeps the data of the call, and sends
+ * one `tool_result` event at the end.
+ *
+ * The CLI has one limit. It sends a web search as `{variant: 'WebSearch',
+ * backend: true}`, with no query and no result, because xAI does the search on
```
mattrossman: Note in my local testing, the result contained queries and results for web searches, e.g.:

Image

packages/core/src/agents/grok/runner.ts
```diff
@@ -0,0 +1,228 @@
...
+
+export const grokRunner: AgentRunner<GrokModel> = {
+  id: 'grok',
+  displayName: 'Grok Build',
+  apiKeyEnvVar: 'XAI_API_KEY',
+  cliPackage: '@xai-official/grok',
+  // Keep this version. The schema of the stream changes between releases, and
+  // the documentation in the CLI does not agree with the events that it sends.
+  // If you increase the version, examine the parser against a true run.
+  defaultCliVersion: '1.0.13',
```
mattrossman: I'll prob try upgrading to latest stable (1.0.34) before merging this in

packages/core/src/agents/grok/runner.ts
```diff
@@ -0,0 +1,228 @@
...
+    return typeof n === 'number' ? n : undefined;
+  },
+
+  deriveStopReason(raw, command) {
+    if (!raw) return processStopReason(command);
+    const { records } = parseJsonlRecords(raw);
+    // The last `end` event has more importance than an earlier `error` event.
+    // Grok also sends an `error` event for a fault that it corrects. Thus an
+    // `error` event before a correct `end` event shows a run that continued.
+    const end = [...records].reverse().find((r) => r.type === 'end');
```
mattrossman: nit, we could use the `lastEndEvent` helper here

packages/core/src/agents/grok/runner.ts
```diff
@@ -0,0 +1,228 @@
...
+    // reads out of `inputTokens`, and the schema of this repository counts them
+    // in `inputTokens`. Thus this function adds them again.
+    //
+    // A run that stops at the time limit sends no `end` event, thus it gives no
+    // data. The `usage` events during the run contain this data. But Claude Code
+    // and Codex also lose the data. Get the data for the three agents, or for no
+    // agent.
+    if (!raw) return undefined;
+    const { records } = parseJsonlRecords(raw);
+    const end = [...records].reverse().find((r) => r.type === 'end');
```
mattrossman: Same as https://github.com/supabase/evals/pull/300/changes#r4049573609, could use `lastEndEvent`

packages/core/src/agents/grok/runner.ts
mattrossman: Tracking in [AI-1233](https://linear.app/supabase/issue/AI-1233/capture-per-request-token-usage)

mattrossman:
> `reasoningEffort: 'high',`

👍 Confirming I see "high" as the default on https://docs.x.ai/developers/grok-4-6#at-a-glance

packages/core/src/agents/grok/parser.ts
mattrossman: We should fix the parser so that `pages: []` gets populated, since it appears to have this data in the raw transcript.

packages/core/src/agents/grok/runner.ts
mattrossman: Bumped the version in https://github.com/supabase/evals/pull/300/changes/72398d8d637dda5f25ff12dad6fd5fba903f0541 and sanity checked no breakage in a sample run.

packages/core/src/agents/grok/runner.ts
mattrossman: Addresed in https://github.com/supabase/evals/pull/300/changes/9dfc244ec48082bdf3d38f83c5ce45bd1780c2f3

packages/core/src/agents/grok/runner.ts
mattrossman: Addressed in https://github.com/supabase/evals/pull/300/changes/9dfc244ec48082bdf3d38f83c5ce45bd1780c2f3

packages/core/src/agents/grok/parser.ts
mattrossman: Updated in https://github.com/supabase/evals/pull/300/commits/c6b9d17e6eac2c937ce436a8878bc314f80e3e40

It's a little tricky b/c Grok uses server-side web page synthesis so I'm leaving `hasContent: unknown` for now. Later on I think we should (A) refactor this docs search parsing stuff into a more normalized pipeline so we don't have harness-specific logic floating around here and (B) make `hasContent: boolean` finer grained into an enum of content types, because there's subtleties for each harness. Sometimes you get a full page, sometimes an LLM summary, or a part of a page, and we need to be able to express those differences.

packages/sandbox/src/skills.ts
mattrossman: Addressed in https://github.com/supabase/evals/pull/300/commits/ef2fb655038683840aa54835a23423373da527fa

mattrossman:
Due to the quirk in [AI-1005](https://linear.app/supabase/issue/AI-1005/support-multiple-explicit-experiment-overrides-on-refresh-workflow) I'm triggering the refresh in 2 phases.

### With skills

```bash
gh workflow run eval-refresh.yml --ref feat/grok-native-harness -f suite=benchmark -f experiment_suite=benchmark -f experiments=grok-4.6 -f runs=3 -f merge=true -f commit_to_branch=true
```

Run: 🟢  https://github.com/supabase/evals/actions/runs/35393677080

Notes:

- I thought `build-functions-005-dual-auth-user-secret` might need a re-run since it hit a 502 from edge functions runtime, which seemed related to the `ensureEdgeRuntime()` workaround from #291, but I'm pretty sure this is just the sandbox timing out from the legitimately slow agent solution

### Without skills

```bash
gh workflow run eval-refresh.yml --ref feat/grok-native-harness -f suite=benchmark -f experiment_suite=no-skills -f experiments=grok-4.6-no-skills -f runs=3 -f merge=true -f commit_to_branch=true
```

Run: 🟢  https://github.com/supabase/evals/actions/runs/35395340881

mattrossman:
Results have shuffled around since the earlier state, Grok no longer on top:

| Benchmark | Without Skills |
|--------|--------|
| CleanShot 2026-09-18 at 5 33 50 PM@2x | CleanShot 2026-09-18 at 5 33 55 PM@2x | 

Looks like the score decrease is mainly because it's taking too long to finish (either due to newer Grok CLI, or newer Supabase CLI) and ends up getting disqualified by timeout, which (I think) is a legit failure reason. It was already timing out a lot on the earlier Grok version so I don't see this as a net new concern. We could consider revisiting `build-functions-005` in a later PR to see why it's more likely to exhaust the time limit on it.

mattrossman (approved):

PR #303, opened 2026-09-16
feat: record sandbox usage on eval results

mattrossman:
Records [Vercel Sandbox's metered usage](https://vercel.com/docs/sandbox/pricing) on eval results so we can estimate total sandbox cost per experiment. See fields mirrored from [`sandbox.stop()` return type](https://vercel.com/docs/sandbox/sdk-reference#sandbox.stop).

```jsonc
/* eval-results.json */
{
    // ...
    "agentRunDurationMs": 12345,
    "sandboxUsage": {
      "activeCpuDurationMs": 490837,
      "duration": 308496,
      "memory": 8192,
      "networkTransfer": {
        "ingress": 2728707529,
        "egress": 4804954
      }
    },
    // ...
}
```

I renamed our `durationMs` -> `agentRunDurationMs` to disambiguate from the new sandbox duration fields.

## Notes

In the previous PR https://github.com/supabase/evals/pull/302 I kept all the files from the Sandbox tarred together to keep the agent-authored ones contained in case of corrupt filenames. This PR moves `result.json` out from that so we can inject the Sandbox usage stats w/o needing to untar / retar the results file.

The runner tests also cover archived agent-authored filenames and incomplete `.partial` results.

## How to review

I sanity checked one benchmark eval via the [refresh workflow](https://github.com/supabase/evals/actions/runs/35152800137). Take a look at the [sandbox usage snapshot](https://github.com/supabase/evals/blob/d9bc1c7447c01b7f015a391a86d0688a43ea8d3d/apps/web/src/data/eval-results.json#L31147-L31155) to vet the new fields.

Closes AI-1222


apps/framework/scripts/export-results.ts
```diff
@@ -162,6 +162,7 @@ async function readResultFile(
     stepCount: parsedResult.stepCount,
     toolCallCount: parsedResult.toolCallCount,
     durationMs: parsedResult.durationMs,
```
mattrossman, line 164: At some point I'd like to rename this to e.g. `agentRunDurationMs` to make it more clear what it's timing, now that we have other durations reported by Sandbox SDK. I left that rename out this PR to reduce noise but could re-add it if others agree.

gregnr: I'm not against this, but also not sure how disruptive it is to rename these. Will defer to yours and others' thoughts

gregnr (approved):

apps/framework/scripts/export-results.ts
mattrossman: Right now the only consumer I know of is our [Evals](https://app.hex.tech/supabase/app/Evals-033abDlwqlTbW5ktgwFffU/latest) Hex (maybe docs team uses it too, checking [here](https://supabase.slack.com/archives/C023E4L60R3/p1789734966409619)) so I see it being simple to update.

mattrossman: I performed the rename in https://github.com/supabase/evals/pull/303/commits/b9182d84f8dd382f6ccecf161a15897c38ded612 and migrated the field in the result snapshots in https://github.com/supabase/evals/pull/303/commits/a51fd5808b77bf4bfed9a2576146b4857c862b19 since I don't see duration consumed by the docs team Hex. After merging I'll update our Hex.

mattrossman:
Hex has been updated with the renamed `durationMs` -> `agentRunDurationMs` field

PR #315, opened 2026-09-21
feat: organize experiments by owner

Rodriguespn:
This change organizes experiments under per-owner folders so CODEOWNERS can scope them by path. It also makes experiment and result discovery recursive for nested folders.

AI-owned base configs live in `experiments/presets.ts`. CLI arms reuse the exported `baselineExperiment`, so the baseline changes in one place.

```text
experiments/
├── presets.ts
├── ai/
└── cli/
    └── _lib/
```

> (Diff note: the changes under `experiments/` are file moves and config extraction. The `apps/web/src/data/*.json` changes only update experiment IDs and source paths. Scores and results are unchanged. The runtime changes to review are the nine files under `apps/framework/`.)

## How to review

Start with `experiment-files.ts`, `result-files.ts`, and `cli-args.ts`. These contain the recursive discovery and nested ID behavior.

Dry run one of the moved experiments:

```bash
pnpm eval:dry -- --eval resolve-dataapi-001-empty-results --experiment ai/claude-code-sonnet-5
```

## Verification

- `pnpm --filter @supabase-evals/framework typecheck`
- `pnpm --filter @supabase-evals/framework test:vercel-runner`
- `pnpm --filter @supabase-evals/framework test:cli-lib`
- `pnpm format:check`
- Nested list, dry-run, and export smoke checks

This PR is stacked on #308.

Context: [Slack thread](https://supabase.slack.com/archives/C051L8U2EJF/p1789748029119999?thread_ts=1789488713.293089&cid=C051L8U2EJF)


apps/framework/lib/experiment-files.ts
```diff
@@ -0,0 +1,37 @@
...
+      owner.name.startsWith('_')
+    ) {
+      continue;
+    }
+
+    const ownerDir = join(experimentsDir, owner.name);
+    for (const entry of await readdir(ownerDir, { withFileTypes: true })) {
+      if (!entry.isFile() || !entry.name.endsWith(EXPERIMENT_SUFFIX)) continue;
+
+      files.push({
```
mattrossman, line 29: In case a team ends up trying to kick off a refresh w/ experiment names that overlap across teams which would silently clash, maybe worth a small duplicate check + error surfaced here.

experiments/presets.ts
```diff
@@ -0,0 +1,84 @@
...
+  agent: codexAgent({
+    model: 'gpt-5.6-luna',
+    reasoningEffort: 'medium',
+  }),
+  runtime: defaultRuntime(),
+  localStack: localStackRuntime(),
+  skills,
+} satisfies ExperimentPreset;
+
+export const baselineExperiment = codexGpt56Luna;
```
mattrossman, line 55: What's the use case for this if codexGpt56Luna is already exported? Assuming this is scaffolding for future PR, maybe better to add once it's used for something.

experiments/ai/claude-code-sonnet-5-no-skills.experiment.ts
```diff
@@ -0,0 +1,8 @@
+import { defineExperiment } from '@supabase-evals/core';
+import { claudeCodeSonnet5 } from '../presets.ts';
```
mattrossman, line 2: I think we'll want to fix the tsconfig btw, IIUC this `.ts` extension would normally fail the typecheck. It look like this was intended to be handled by the `"experiments"` includes entry in `apps/framework/tsconfig.json` but rn that resolves to `apps/framework/experiments` instead of root `experiments/`

experiments/presets.ts
mattrossman: If the intention was to signal to other teams which base they should use—perhaps the other harness/model combos shouldn't be in `presets.ts`? Unless we know other teams want to use them. For now it seems like other teams will only be using 5.6 Luna.

.github/CODEOWNERS
```diff
@@ -1,4 +1,7 @@
 * @supabase/ai
 
+/experiments/ai/ @supabase/ai
+/experiments/presets.ts @supabase/ai
```
mattrossman, line 4: nit, these are already handled by the `@supabase/ai` catch-all

mattrossman:
Sanity checking w/ a regression run to verify results look normal still:
https://github.com/supabase/evals/actions/runs/35649310761

mattrossman (approved):

mattrossman:
The parts of the refresh that worked look fine, the broken parts are my fault from a bug in #303 that slipped in 🙈 will fix that separately.

(Edit: fixed the regression in https://github.com/supabase/evals/pull/321 so you may wanna merge that in here to fully confirm regression results still look normal)

experiments/presets.ts
Rodriguespn: > signal to other teams which base they should use

This was the main motivation yes. From the other team's POV, this means that they only have to import the baseline model. If we change the baseline model (like we recently did from gpt5.4-mini to 5.6-luna, we don't need to find every place where the baseline model is and change it

apps/framework/lib/experiment-files.ts
Rodriguespn: Good call. Duplicated experiment names now throw an error on collision

experiments/presets.ts
Rodriguespn: Thinking about this again, each team should choose the config for its experiments. We can steer teams toward a shared default later if that becomes useful.

Rodriguespn:
Re-ran the regression suite after rebasing onto #321. All 78 jobs completed and published successfully, with no packaging or discovery errors. The benchmark result was 50/78 passes, compared with 58/78 previously. The refreshed results are in https://github.com/supabase/evals/pull/319.

Merging this 🚢

